Issue #47, opened 2020-03-06
Add new ReEDS supply curve binning method: "equal_cap"


mmowers:
**Why this feature is necessary:**
kmeans is the current binning method, but it has problems with reproducibility, ordering of bins, and amount of capacity in each bin. This method will ensure that approximately equal amounts of capacity are in each supply curve bin for a given region/class.

**A possible solution is:**
For a given region/class, calculate total capacity of resource and divide by number of bins to get the bin size. Then sort the supply curve by trans_cap_cost for this region/class. Iterate through the rows of the supply curve and, when total cumulative capacity exceeds the bin size, the next point will go into the next bin.

**I have considered the following alternatives:**
Jenks natural breaks, a 1-D version of kmeans. But the problem with this (as with kmeans) is that they don't consider the amount of capacity at each supply curve point.

**Urgency / Timeframe**
To be discussed 2020-03-09

mmowers:
Actually, there’s a problem with this. Let’s say we have 5 bins and 6 supply curve points. But let’s say the most expensive of the supply curve points has amounts all the capacity, e.g. 90% of it. In this case we might put all the other capacity into the first bin. Let's hold off on implementation until we figure this out. I might test out jenks or some other alternative in the meantime.

MRossol:
@mmowers I think the above is an EXTREME fringe case and not worth abandoning this approach.
The SC points have a finite capacity based on the aggregation size (area) and power density (currently constant by technology).  Therefore the dispersion in capacities between SC points will be fairly small. Furthermore, you'd have to use an EXTREMELY small geo-spatial region to have the same order of SC points as bins.

I'll implement this today before out meeting and add in a warning IF (again big if) the maximum SC capacity is > 1/2 the bin size.

MRossol:
@mmowers 
Follow up question, do you want the same NUMBER of equal size bin in each "region"-"class", or do you want all bins to be of a given SIZE.  I'm going to build in the former for now, we can discuss this PM.

mmowers:
I've checked the equal capacity bins and they generally look good to me. Even if we see a few instances of the bin crowding issue i raised above, I think we're good to go with this method. Thanks!

MRossol:
Great. I did notice that in some cases bins are not filled in order. I can easily fix that if it would be helpful to always have bins filled in order.  The only upside I can see to not is that you know where in the range of total capacity each site sites.

Let me know either way

mmowers:
[...]

PR #48, opened 2020-03-09
replace clusters_bins with capacity_bins

MRossol:
Implementation of #47 
Replace cluster_sc_bins with capacity_bins
- For each "resource"-"class" group create the given number of equal capacity bins, by default 3
- Fill bins with cumulative capacity after sorting by the given column(s),  by default "trans_cap_cost"

Misc:
- config class updated
- reeds_cli updated
- example configs updated
- tests updated


MRossol:
fixes #50

PR #57, opened 2020-04-06
Mnr/sc powerrose

MRossol:
Add prominent wind direction computation sub-package, cli and config


MRossol:
@grantbuster  Need to make tests, but feel free to start reviewing the class in the mean time

reVX/wind_dirs/prominent_wind_dirs.py
```diff
@@ -0,0 +1,244 @@
...
+        neighbor_gids = pd.DataFrame(neighbor_gids,
+                                     columns=directions,
+                                     index=sc_point_gids)
+
+        return neighbor_gids
+
+    def _get_prominent_directions(self, excl_area=0.0081, max_workers=None,
+                                  chunk_point_len=100):
+        """
+        Aggregate power rose data to
```
grantbuster, line 144: You've to me on the edge of my seat!

reVX/wind_dirs/prominent_wind_dirs.py
```diff
@@ -0,0 +1,244 @@
...
+            agg_dset='powerrose_100m', tm_dset='techmap_wtk',
+            resolution=64, excl_area=0.0081, max_workers=None,
+            chunk_point_len=100, out_fpath=None):
+        """
+        Aggregate powerrose to supply curve points, find neighboring supply
+        curve point gids and rank them based on prominent powerrose direction
+
+        Parameters
+        ----------
+        power_rose_h5_fpath : str
```
grantbuster, line 201: This could benefit from some description of the data structure of the power rose dataset. For example, columns are sites and rows are directions? Or vis versa?

reVX/wind_dirs/prominent_wind_dirs.py
```diff
@@ -0,0 +1,244 @@
...
+        Returns
+        -------
+        neighbor_gids : ndarray
+            Neighboring supply curve point gids
+        """
+        sc_points = points.loc[sc_point_gids]
+        rows, cols = sc_points.values.T
+        shape = points['row_ind'].max(), points['col_ind'].max()
+
+        row_shifts = [-1, -1, 0, 1, 1, 1, 0, -1]
```
grantbuster, line 88: Can you explain this to me?

reVX/wind_dirs/prominent_wind_dirs.py
```diff
@@ -0,0 +1,244 @@
...
+        chunk_point_len : int, optional
+            Number of SC points to process on a single parallel worker,
+            by default 100
+        out_fpath : str
+            Path to .csv file to save output table to
+
+        Returns
+        -------
+        prominent_dirs : pandas.DataFrame
+            Update meta data table with neighboring supply curve point gids
```
grantbuster, line 229: For me at least this is kind of nebulous. How are the SC point gids stored? Are there 8 new columns, one for each direction? What is the order of directions (see other comment on direction attr).

reVX/wind_dirs/prominent_wind_dirs.py
```diff
@@ -0,0 +1,244 @@
...
+            Neighboring sc_point_gids by cardinal direction
+        """
+        with SupplyCurveExtent(excl_fpath, resolution=resolution) as sc:
+            points = sc.points
+
+        neighbor_gids = \
+            ProminentWindDirections._compute_neighbors(sc_point_gids,
+                                                       points)
+
+        directions = ['N', 'NE', 'E', 'SE', 'S', 'SW', 'W', 'NW']
```
grantbuster, line 134: This might be a good class attribute so people could easily see what the global direction order is.

reVX/wind_dirs/prominent_wind_dirs.py
MRossol: I swear I already wrote that...

reVX/wind_dirs/prominent_wind_dirs.py
MRossol: done, its WTK with directions on axis 0

reVX/wind_dirs/prominent_wind_dirs.py
MRossol: Which part? I replaced lines 84 -> 86 with a new method that takes the sc_point_gids and shape and computes the row and col indices.

The rest of the method computes the gids from the row and column indices for all of the neighbors

reVX/wind_dirs/prominent_wind_dirs.py
MRossol: tried to be more verbose, what I am currently doing is adding 16 new columns:
8 = the cardinal direction of each neighbor and its gid
8 = the same gids sorted in terms of prominent power rose direction

while these are redundant I thought it might be nice to have both

reVX/wind_dirs/prominent_wind_dirs.py
MRossol: Done

PR #60, opened 2020-07-06
PLEXOS Plants

MRossol:
Feature to build and aggregate Plexos plants from supply curve points


MRossol:
WFED 11578 01.01.04

reVX/handlers/sc_points.py
```diff
@@ -0,0 +1,618 @@
...
+            msg = ('Must supply either a jsonified list or a list, not {}'
+                   .format(type(list_in)))
+            logger.error(msg)
+            raise ValueError(msg)
+
+        out = np.array(list_in, dtype=dtype)
+
+        return out
+
+    def _drop(self, build_capacity, drop=None):
```
grantbuster, line 156: This is a confusing method... Consider renaming to build_capacity or drop_built_capacity. Additionally, the docstring is incomplete (add description of build_capacity parameter). 

On a technical note, this always builds out the first gids in the resource gid list right? What kind of guarantee is there that these are the gids that should actually be built? I think i see the sorting happening in __init__, but this should be described in the doc string here (build the resource gids with the best capacity factor)

reVX/handlers/sc_points.py
```diff
@@ -0,0 +1,618 @@
...
+        sc_point = pd.Series(sc_point)
+
+        if drop is None:
+            self._res_gids = None
+            self._gid_counts = None
+            self._res_capacity = None
+            self._cf_means = None
+            self._capacity = 0.0
+        else:
+            self._res_gids = np.delete(self._res_gids, drop_slice)
```
grantbuster, line 193: Why wouldnt you just set the capacity to zero for the "dropped" resource gid? I think setting available capacity to zero makes more sense than removing array indices.

reVX/handlers/sc_points.py
```diff
@@ -0,0 +1,618 @@
...
+        Returns
+        -------
+        out : tuple
+            (sc_point, capacity, availability)
+        """
+        build_capacity = capacity.copy()
+        if self.capacity > 0:
+            if capacity < self.capacity:
+                drop = 0
+                for cap in self.resource_capacity:
```
grantbuster, line 224: I'm surprised that this loop works in conjunction with np.delete() in _drop(). Doesn't np.delete() remove an item from the array? Wouldn't continually removing items from the array and incrementing the drop index result in a broken iteration? see my comment on setting available capacity to zero instead of np.delete().

reVX/handlers/sc_points.py
```diff
@@ -0,0 +1,618 @@
...
+        Extract capacity from Supply Curve point
+
+        Parameters
+        ----------
+        capacity : float
+            Capacity to extract from Supply Curve point
+
+        Returns
+        -------
+        out : tuple
```
grantbuster, line 217: You should copy the docstring from _drop(). Better to have the descriptive doc string on the public method!

reVX/plexos/plexos_cli.py
```diff
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 """
-PLEXOS command line interface (cli).
+reV-ReEDS-PLEXOS command line interface (cli).
```
grantbuster, line 3: maybe rename this file to rev_reeds_plexos_cli.py? Also I think cli should be a prefix when possible.

reVX/plexos/plexos_plants.py
```diff
@@ -0,0 +1,1006 @@
...
+            LCOE column to sort by, by default 'total_lcoe'
+        lcoe_thresh : float, optional
+            LCOE threshold multiplier, exclude sc_gids above threshold,
+            by default 1.3
+        plants_per_worker : int, optional
+            Number of plants to identify on each worker, by default 40
+
+        Returns
+        -------
+        plants : list
```
grantbuster, line 650: Is this actually supposed to return something? Doesn't look like it does

reVX/plexos/plexos_plants.py
```diff
@@ -0,0 +1,1006 @@
...
+            by default 1.3
+        max_workers : int, optional
+            Number of workers to use for plant sc extraction, 1 == serial,
+            > 1 == parallel, None == parallel using all available cpus,
+            by default None
+
+        Returns
+        -------
+        plants : list
+            List of plant supply curve tables
```
grantbuster, line 604: Is this actually supposed to return something? Doesn't look like it does

reVX/plexos/plexos_plants.py
```diff
@@ -0,0 +1,1006 @@
...
+        max_workers : int, optional
+            Number of workers to use for plant identification, 1 == serial,
+            > 1 == parallel, None == parallel using all available cpus,
+            by default None
+        plants_per_worker : int, optional
+            Number of plants to identify on each worker, by default 40
+
+        Returns
+        -------
+        plants : list
```
grantbuster, line 428: I think this could be more descriptive... I think i know what this is but only because we talked through all the nonsense required to associate SC points with plexos plants. I think I'll have no idea what this is in a few months.

reVX/plexos/plexos_plants.py
```diff
@@ -0,0 +1,1006 @@
...
+
+        # Sort by distance
+        plant_sc['dist'] = PlexosPlants._haversine_dist(plant_coords,
+                                                        sc_coords)
+        plant_sc = plant_sc.sort_values('dist')
+
+        return plant_sc.reset_index(drop=True)
+
+    @staticmethod
+    def _identify_plants(plant_table, sc_table, dist_percentile=90,
```
grantbuster, line 403: I think this algorithm could be its own class. Very unique purpose when compared to "building" the plant profiles. It's also confusing having this next to methods like "_allocate_sc_gids()" which sound like they could be for identifying / associating SC points with the Plexos plants.

reVX/plexos/plexos_plants.py
```diff
@@ -0,0 +1,1006 @@
...
+                    else:
+                        idxs = np.argsort(sc_dists)
+
+                    plant_ids = plant_ids[idxs]
+
+                for plant_id in plant_ids:
+                    capacity = self.plant_capacity[plant_id]
+                    if (capacity > 0) and self.sc_points.check_sc_gid(sc_gid):
+                        sc_point, sc_capacity = \
+                            self.sc_points.get_capacity(sc_gid, capacity)
```
grantbuster, line 562: doesnt get_capacity return three items?

reVX/plexos/plexos_plants.py
```diff
@@ -0,0 +1,1006 @@
...
+                                plant = [sc_point]
+                            else:
+                                plant.append(sc_point)
+
+                            self[plant_id] = plant
+                            self._capacity[plant_id] -= sc_capacity
+                            logger.debug('Allocating {:.1f}MW to plant {} from'
+                                         ' sc_gid {}'.format(sc_capacity,
+                                                             plant_id,
+                                                             sc_gid))
```
grantbuster, line 575: I find this wildly confusing.... Is plant a PlexosPlant object? Why are you appending an SC point to the plant? I think in reality the plant is a list of SC points associated with each plant, but then why is it called plant? Maybe we should break out the "plant" into a standalone object.

reVX/plexos/plexos_plants.py
```diff
@@ -0,0 +1,1006 @@
...
+                        profile += cf_profile * gid_capacities[i]
+
+        if len(profile.shape) != 1:
+            profile = profile.flatten()
+
+        return profile
+
+    def aggregate_plants(self, out_fpath):
+        """
+        Aggregate plants from capacity factor profiles and save to given
```
grantbuster, line 787: You're really aggregating the profiles right? Making aggregated power profiles from the source resource profiles once the "allocation" step is done? The naming and descriptions are confusing. In a vacuum, I would have no idea when to use this method.

reVX/plexos/plexos_plants.py
```diff
@@ -0,0 +1,1006 @@
...
+    def _make_profile(cf_fpath, plant_build):
+        """
+        Make generation profiles for given plant buildout
+
+        Parameters
+        ----------
+        cf_fpath : str
+            Path to reV Generation output .h5 file to pull CF profiles from
+        plant_build : pandas.DataFrame
+            DataFrame describing plant buildout
```
grantbuster, line 761: Can you describe this description :) ?

Is this the list of SC points associated with this plant? With a capacity column maybe?

reVX/handlers/sc_points.py
```diff
@@ -0,0 +1,618 @@
...
+        sc_gid = sc_point.name
+        capacity = sc_point['capacity']
+        res_gids = sc_point['res_gids']
+        gid_counts = sc_point['gid_counts']
+        res_cf = res_cf_means.loc[res_gids].values
+
+        return cls(sc_gid, capacity, res_gids, gid_counts, res_cf)
+
+    @staticmethod
+    def _create_worker_slices(table, points_per_worker=400):
```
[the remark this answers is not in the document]
MRossol: moved

reVX/handlers/sc_points.py
MRossol: renamed and updated

reVX/handlers/sc_points.py
MRossol: added a new list to modify called _avail_cap

reVX/handlers/sc_points.py
MRossol: Remove np.delete

reVX/handlers/sc_points.py
MRossol: updated

reVX/plexos/plexos_cli.py
MRossol: renamed, though I like cli at the end...

reVX/plexos/plexos_plants.py
MRossol: nope old docstring

reVX/plexos/plexos_plants.py
MRossol: ditto

reVX/plexos/plexos_plants.py
MRossol: done

reVX/plexos/plexos_plants.py
MRossol: renamed the same method from Point to avoid confusion. This method is on SupplyCurvePoint and only returns two values

reVX/handlers/sc_points.py
```diff
@@ -0,0 +1,631 @@
...
+            (sc_point, capacity, availability)
+        """
+        build_capacity = capacity.copy()
+        if self.capacity > 0:
+            if capacity < self.capacity:
+                drop = 0
+                for cap in self.available_capacities:
+                    drop += 1
+                    capacity -= cap
+                    if capacity <= 0:
```
grantbuster, line 239: Do you always get even builds from this? Like can it build part of a resource gid? I would think the build from each resource gid would be boolean right? This is also related to this line 212: `self._avail_cap[drop_slice] = 0.0`

reVX/handlers/sc_points.py
```diff
@@ -0,0 +1,631 @@
...
+                                  offshore=offshore)
+
+    def __repr__(self):
+        msg = "{} with {} points".format(self.__class__.__name__, len(self))
+        return msg
+
+    def __len__(self):
+        return len(self._sc_points)
+
+    def __getitem__(self, sc_gid):
```
grantbuster, line 315: I really think the getters and setters should have docstrings. Just because pylint doesnt require it doesnt mean its not important. What does getting an sc_gid from this object return? No idea.

reVX/plexos/plexos_plants.py
```diff
@@ -0,0 +1,1113 @@
...
+        pp = cls(plexos_table, sc_table, cf_fpath,
+                 dist_percentile=dist_percentile, lcoe_col=lcoe_col,
+                 lcoe_thresh=lcoe_thresh, offshore=offshore,
+                 max_workers=max_workers, plants_per_worker=plants_per_worker,
+                 points_per_worker=points_per_worker)
+
+        pp.dump(out_fpath=out_fpath)
+
+
+class PlantProfileAggregation:
```
grantbuster, line 668: I think this division of classes makes a lot more sense. I still think you should have used the pre-existing PlexosNode object but thats okay.

reVX/handlers/sc_points.py
MRossol: While I allow the splitting of a Supply Curve point I only let each Resource gid to be associated with a single "Plant". The logic is that we extract capacity from the resource gids until we have (over)filled the desired capacity. This is all a bit redundant as in the end we multiply the resource CF profiles by the actual desired build capacity and not the actual capacity available at each resource gid... 

For line 212, this is how I'm tracking which resource gids have been "allocated" to plants. It replaces the old np.delete logic.

reVX/handlers/sc_points.py
MRossol: Fair enough, added

reVX/plexos/plexos_plants.py
MRossol: This was more difficult than it seems. The only bit of code that is really shared with PlexosNode is in this method:
https://github.com/NREL/reVX/blob/423e0abf7614f73639ebaa4e57ce1a30b82c377f/reVX/plexos/plexos_plants.py#L962
But that is embedded within the PlexosNode aggregation logic here: 
https://github.com/NREL/reVX/blob/423e0abf7614f73639ebaa4e57ce1a30b82c377f/reVX/plexos/rev_reeds_plexos.py#L162

The rest is unique to the Plexos Plant input.

I love the idea of generalizing some of this, but its tough with how unique (and different) the inputs are from ReEDS vs Plexos.  Down the road, we could try to generalize this profile making bit and maybe the output code to take any set of datasets.

reVX/handlers/sc_points.py
grantbuster: Yeah but say you have three resource gids with capacities [1, 1, 8] and you're building a plant with capacity 3. Would the final profile be weighted 80% to the third resource gid even though all three should be weighted equally? Or does that get handled somewhere else?

MRossol: That will get handled by the Aggregation class. Here is the high level logic:
When sc_gids are "allocated" to a plant their "capacities" are recomputed to sum to the desired plant build capacity.
E.g.
Plant 1 needs 100MW which will be allocated from sc_gid 1 and 2 (in that order)
sc_gid 1 has a capacity of 50MW -> all 50MW are "built"
sc_gid 2 has a capacity of 75MW -> only 50MW will be "built" as plant 1 only needs 50MW of capacity from sc_gid 2

During aggregation the resource gids capacities are computed from the gid_counts and the sc_gid build capacity.
e.g. lets say sc_gid 2 has 2 res_gids with capacities [25, 50] (recall the actual capacity of sc_gid 2 was 75MW).  But only 50MW is being built, so the res_capacities are recomputed as [17, 33].  So the distribution of capacities is retained.  

I realize that in both cases this isn't "perfect" as the largest res_gid is the last one, but on the fly changing the gid_counts or tracking the res_capacity seemed like overkill. Maybe a potential improvement down the road...

PR #68, opened 2020-11-25
Forecast corrector

MRossol:
@grantbuster Thoughts on data to use for tests?


grantbuster:
Synthetic data? :)

reVX/utilities/forecasts.py
```diff
@@ -0,0 +1,386 @@
...
+
+        Returns
+        -------
+        agg_mae : float
+            Aggregate MAE for all sites
+        ave_mae : float
+            Average MEA for all sites
+        """
+        agg_mae = (np.abs(np.nansum(fcsts - actuals, axis=1)).sum()
+                   / np.nansum(np.nanmax(actuals, axis=0) * 8760))
```
grantbuster, line 199: 8760 hard coded? Can't you just use len()?

reVX/utilities/forecasts.py
```diff
@@ -0,0 +1,386 @@
...
+        ----------
+        actuals : ndarray
+            Timeseries actuals (time x sites)
+        fcsts : ndarray
+            Timeseries forecats (time x sites)
+
+        Returns
+        -------
+        agg_mae : float
+            Aggregate MAE for all sites
```
grantbuster, line 194: What is this? The sum of the MAE for all sites?

reVX/utilities/forecasts.py
```diff
@@ -0,0 +1,386 @@
...
+            Timeseries forecats (time x sites)
+        fcst_perc : float
+            Percentage of forecast to use for blending
+
+        Returns
+        -------
+        fcsts : ndarray
+            Bias corrected and blended forecasts
+        """
+        fcsts = cls.bias_correct_fcst(actuals, fcsts)
```
grantbuster, line 137: I think you should include in both the method doc strings and the class doc string the exact steps taken to manipulate the forecast data. This will help with transparent documentation propagated through to the docs.

grantbuster (changes requested):
Looks good. Few minor comments. I think tests with synthetic data would be effectve.

reVX/utilities/forecasts.py
MRossol: whoops, took this from Greg, good catch

reVX/utilities/forecasts.py
MRossol: As I remember yes. Look at the over all MAE for all sites vs the average MAE across sites

reVX/utilities/forecasts.py
MRossol: Good call

reVX/utilities/forecasts.py
```diff
@@ -200,10 +201,10 @@ def compute_mae(actuals, fcsts):
             Average MEA for all sites
         """
         time_steps = len(fcsts)
-        agg_mae = (np.abs(np.nansum(fcsts - actuals, axis=1)).sum()
-                   / np.nansum(np.nanmax(actuals, axis=0) * time_steps))
-        ave_mae = np.nanmean(np.abs(fcsts - actuals).sum(axis=0)
-                             / (actuals.max(axis=0) * time_steps))
+        site_mae = np.mean(np.abs(fcsts - actuals), axis=0)
+        site_rel_mae = site_mae / (np.max(actuals, axis=0) * time_steps)
```
grantbuster, line 205: wait do we actually need to divide by the number of steps anymore?

grantbuster (approved):

Issue #65, opened 2020-11-16
Exclusion Setbacks


MRossol:
**Why this feature is necessary:**
Need to develop spatial setbacks for reV as an exclusions layer.
options:
1) Build setbacks using given data (.csv)
2) Build global setbacks using setback model and multiplier

Inputs:
1) hub_height
2) Rotor_diameter
3) microsoft building database (geojson)
    * tip_height (computed from hub_height and rotor_diameter)
4) .csv with existing setbacks
5) optional multiplier
6) optional setback method
7) Wind class mapping (TBD)

**Charge code**
Lopez to provide

**Urgency / Timeframe**
Lopez to provide

PR #69, opened 2020-12-08
Wind set backs

MRossol:
Implementation of #65 
TODO:
CLI
tests


reVX/wind_setbacks/wind_setbacks.py
```diff
@@ -0,0 +1,1386 @@
...
+    def compute_setbacks(self, structure_fpath, max_workers=None,
+                         layer=None, description=None, replace=False):
+        """
+        Compute structure setbacks for all states either in serial or parallel.
+        Existing setbacks are computed if a wind regulations file was supplied
+        during class initialization, otherwise general setbacks are computed
+
+        Parameters
+        ----------
+        features_dir : str
```
grantbuster, line 793: great doc strings but incorrect description. probably elsewhere too. 

What kind of files should these be? What extension does the code look for?

reVX/wind_setbacks/wind_setbacks.py
```diff
@@ -0,0 +1,1386 @@
...
+        setbacks = cls(excl_h5, hub_height, rotor_diameter,
+                       regs_fpath=regs_fpath, multiplier=multiplier,
+                       hsds=False, chunks=chunks)
+        setbacks.compute_setbacks(features_fpath, layer=layer_name,
+                                  max_workers=max_workers,
+                                  description=description,
+                                  replace=replace)
+
+
+class RailWindSetbacks(TransmissionWindSetbacks):
```
grantbuster, line 1337: why are rail setbacks based on transmission while the others are not? Is it that you can't build turbines within x distance of structures and roads, but you can't build transmission lines within x distance of railways? If so, how are we going to use the rail wind setbacks for our non-existent transmission code? Or is this a grue problem?

reVX/wind_setbacks/wind_setbacks.py
```diff
@@ -0,0 +1,1386 @@
...
+            Flag to replace existing layer data with arr if layer already
+            exists in the exlcusion .h5 file, by default False
+
+        Returns
+        -------
+        setbacks : ndarray
+            Raster array of setbacks, ready to be written to the exclusions
+            .h5 file as a new exclusion layer
+        """
+        setbacks = super().compute_setbacks(structure_fpath, groupby='State',
```
grantbuster, line 817: I'm of the *opinion* that subclass methods that only call the inherited method via super() with a kwarg change should not be subclass methods. It's just not worth the extra code due to additional complexity and maintenance. Why not remove this subclass method and call `BaseWindSetbacks.compute_setbacks()` with the kwarg in the `subclass.run()` methods?

reVX/wind_setbacks/wind_setbacks.py
```diff
@@ -0,0 +1,1386 @@
...
+    """
+    MULTIPLIERS = {'high': 3, 'moderate': 1.1}
+
+    def __init__(self, excl_h5, hub_height, rotor_diameter, regs_fpath=None,
+                 multiplier=None, hsds=False, chunks=(128, 128)):
+        """
+        Parameters
+        ----------
+        excl_h5 : str
+            Path to .h5 file containing or to contain exclusion layers
```
grantbuster, line 38: This is where the new setbacks will be written? This is not a source file, correct? If it is a source file, what is being used from it?

reVX/wind_setbacks/wind_setbacks.py
```diff
@@ -0,0 +1,1386 @@
...
+            blade tip height which is used to determine setback distance
+        rotor_diameter : float | int
+            Turbine rotor diameter (m), used along with hub height to compute
+            blade tip height which is used to determine setback distance
+        regs_fpath : str | None, optional
+            Path to wind regulations .csv file, if None create global
+            setbacks, by default None
+        multiplier : int | float | str | None, optional
+            setback multiplier to use if wind regulations are not supplied,
+            if str, must one of {'high': 3, 'moderate': 1.1},
```
grantbuster, line 50: if str must be one of those keys? Kind of unclear. Maybe say that `if str must be a key in cls.MULTIPLIERS`?

Also what is the multiplier used for? I think based on our conversations and my fuzzy memory that this multiplies the max blade tip height to determine the setback, but that would be worthwhile to include in this description.

reVX/wind_setbacks/wind_setbacks.py
```diff
@@ -0,0 +1,1386 @@
...
+                             .format((i + 1), len(features)))
+
+        setbacks = self._rasterize_setbacks(setbacks)
+
+        return setbacks
+
+    def _existing_setbacks(self, features_state_map, groupby='State',
+                           max_workers=None):
+        """
+        Compute existing setbacks for all states either in serial or parallel
```
grantbuster, line 520: What does this do? The exclusions with setbacks are already in a shape file and you're just rasterizing them? Digging into the code i think this is calculating setbacks from pre-existing structures with pre-existing state setback regulations, but that's not how this reads to me.

reVX/wind_setbacks/wind_setbacks.py
```diff
@@ -0,0 +1,1386 @@
...
+
+        Parameters
+        ----------
+        features_state_map : list
```
grantbuster, line 524: bad docstring

reVX/wind_setbacks/wind_setbacks.py
```diff
@@ -0,0 +1,1386 @@
...
+        for i in range(len(wind_regs)):
+            cnty = wind_regs.iloc[[i]]
+            setback = cls._get_setback(cnty.iloc[0], tip_height,
+                                       rotor_diameter)
+            if setback is not None:
+                logger.debug('- Computing setbacks for county FIPS {}'
+                             .format(cnty.iloc[0]['FIPS']))
+                tmp = gpd.sjoin(features, cnty,
+                                how='inner', op='intersects')
+                tmp['geometry'] = tmp.buffer(setback)
```
grantbuster, line 379: ah this is where the magic happens! sweet.

reVX/wind_setbacks/wind_setbacks.py
```diff
@@ -0,0 +1,1386 @@
...
+            run in parallel on all available cores, by default None
+
+        Returns
+        -------
+        setbacks : ndarray
+            Raster array of setbacks, ready to be written to the exclusions
+            .h5 file as a new exclusion layer
+        """
+        if max_workers is None:
+            max_workers = os.cpu_count()
```
grantbuster, line 483: if you just do `if max_workers == 1: serial else: parallel` and send max_workers to the process pool directly with None/int you don't need to retrieve the os.cpu_count(). Obviously you can't do `max_workers > 1` if `max_workers is None` but you can do `max_workers == 1` if `max_workers is None`, Yay dynamic typing!

reVX/wind_setbacks/wind_setbacks.py
```diff
@@ -0,0 +1,1386 @@
...
+        -------
+        regs : geopands.GeoDataFrame | None
+            Wind Regulations
+        Multiplier : float | None
+            Generic setbacks multiplier
+        """
+        if regs_fpath:
+            if multiplier:
+                msg = ('A wind regulation .csv file was also provided and '
+                       'will be used to determine setback multipliers!')
```
grantbuster, line 414: while this is clear in the source code, it would not be clear to a user what the csv was provided in addition to (the multiplier).

reVX/wind_setbacks/wind_setbacks.py
```diff
@@ -0,0 +1,1386 @@
...
+
+        Returns
+        -------
+        trans : geopandas.GeoDataFrame.sindex
+            Geometries for transmission features, in exclusion coordinate
+            system
+        """
+        trans = gpd.read_file(transmission_fpath)
+        mask = (~(trans['Proposed'] == 'Proposed')
+                & ~(trans['Undergroun'] == 'T'))
```
grantbuster, line 1154: Undergroun a typo? (missing the d)

Also this is a super weird mask. Why not use `!=` instead of `~==`?

reVX/wind_setbacks/wind_setbacks.py
```diff
@@ -0,0 +1,1386 @@
...
+                       hsds=False, chunks=chunks)
+        setbacks.compute_setbacks(roads_dir, layer=layer_name,
+                                  max_workers=max_workers,
+                                  description=description,
+                                  replace=replace)
+
+
+class TransmissionWindSetbacks(BaseWindSetbacks):
+    """
+    Transmission Wind setbacks
```
grantbuster, line 1110: without having to do a code diff can you explain how/why some of these transmission setbacks and methods are different than the structure setbacks?

grantbuster (changes requested):
Really nice code organization, great example of refactoring a notebook into something coherent. I do have a ton of questions / suggestions on how the docs could be improved. This is both because of my ignorance and also i think its crucial for posterity.

Issue #83, opened 2021-01-29
Advance wind/solar temporal stats


MRossol:
**Why this feature is necessary:**
Need to be able to compute temporal statistics from coincident wind and solar generation

**A possible solution is:**
- Subclass rex TemporalStats
- Add ability to read multiple source files (e.g., wind and solar gen)
- Need to add daily stats
- Add in Atlas' new advanced stats methods:
   1. Pearson
   2. Cross-correlation
   3. Stability Coefficient
   4. Spearman
   5. Kendall

**Charge code**
GMLC 10246 09.05.03

**Urgency / Timeframe**
2/12/2021

MRossol:
https://iopscience.iop.org/article/10.1088/1748-9326/aad8f6/pdf

MRossol:
Cross-correlation:
https://onlinelibrary.wiley.com/doi/pdf/10.1002/we.2095

MRossol:
implemented in #89

PR #89, opened 2021-02-10
Hybrid stats

MRossol:


reVX/hybrid_stats/hybrid_stats.py
```diff
@@ -0,0 +1,829 @@
...
+        """
+        Parameters
+        ----------
+        solar_h5 : str
+            Path to solar h5 file(s)
+        wind_h5 : str
+            Path to wind h5 file(s)
+        statistics : str | tuple | dict, optional
+            Statistics to extract, either a key or tuple of keys in
+            cls.STATS, or a dictionary of the form
```
grantbuster, line 35: Where is cls.STATS? I dont see it defined in this class?

reVX/hybrid_stats/hybrid_stats.py
```diff
@@ -0,0 +1,829 @@
...
+
+    @staticmethod
+    def _format_grp_names(index):
+        """
+        Format groupby index values
+
+        Parameters
+        ----------
+        index : list
+            Group by index values
```
grantbuster, line 347: This whole thing is confusing to me. What index is this?

reVX/hybrid_stats/hybrid_stats.py
```diff
@@ -0,0 +1,829 @@
...
+        ----------
+        solar_h5 : str
+            Path to solar h5 file(s)
+        wind_h5 : str
+            Path to wind h5 file(s)
+        sites : pandas.DataFrame
+            Subset of meta DataFrame with sites to extract
+        statistics : dict
+            Dictionary of statistic functions/kwargs to run
+        dataset : str
```
grantbuster, line 477: Would you want to open this up to a solar_dataset and wind_dataset so you could do it on the Resource data if necessary?

reVX/hybrid_stats/hybrid_stats.py
```diff
@@ -0,0 +1,829 @@
...
+            Statistics to extract, either a key or tuple of keys in
+            cls.STATS, or a dictionary of the form
+            {'stat_name': {'func': *, 'kwargs: {**}}}
+
+        Returns
+        -------
+        stats : dict
+            Dictionary of statistic functions/kwargs to run
+        """
+        STATS = {'pearson': {'func': self._pearson_correlation},
```
grantbuster, line 605: Make this a class attribute? Would be nice to be able to easily see what stats are available.

reVX/hybrid_stats/hybrid_stats.py
```diff
@@ -0,0 +1,829 @@
...
+        """
+         Statistics to extract, either a key or tuple of keys in
+        cls.STATS, or a dictionary of the form
+        {'stat_name': {'func': *, 'kwargs: {**}}}
+
+        Parameters
+        ----------
+        statistics : dict
+        """
+        self._stats = self._check_stats(statistics)
```
grantbuster, line 99: I feel like self._check_stats() could just be put in this property setter since its the only place its called.

grantbuster (approved):

reVX/hybrid_stats/hybrid_stats.py
MRossol: I modified this code from TemporalStats, where I do just this. BUT i wasn't sure how to do that since in this case the functions are staticmethods. I guess I could make them pure functions outside of the class

reVX/hybrid_stats/hybrid_stats.py
MRossol: Good call, will do

reVX/hybrid_stats/hybrid_stats.py
MRossol: See below, its a hold over, based on how we decide to handle the default STATS I'll fix the docstring

reVX/hybrid_stats/hybrid_stats.py
MRossol: I'll re-name, index is the groupby names. I was actually proud of this code, but yes it is confusing as F...
The goal is to properly convert the group names into year-month-day-hour (at most) that are all unique

reVX/hybrid_stats/hybrid_stats.py
```diff
@@ -0,0 +1,1588 @@
...
+        corr : ndarray
+            Cross-correlation coefficient for each solar, wind site pair
+        """
+        solar_u = solar_data.mean(axis=0)
+        solar_s = solar_data.std(axis=0)
+        solar_data = np.abs((solar_data - solar_u) / solar_s)
+
+        wind_u = wind_data.mean(axis=0)
+        wind_s = wind_data.mean(axis=0)
+        wind_data = np.roll(wind_data, m, axis=0)
```
grantbuster, line 1001: Why do you roll one of the datasets? Is this to correlate daytime solar with 12-hour offset nighttime wind (for example)?

reVX/hybrid_stats/hybrid_stats.py
```diff
@@ -0,0 +1,1588 @@
...
+        hybrid_stats = cls(solar_h5, wind_h5, res_cls=res_cls)
+        out_stats = hybrid_stats.compute_stats(
+            'cf_profile', month=month, combinations=combinations,
+            reference=reference, max_workers=max_workers,
+            sites_per_worker=sites_per_worker,
+            lat_lon_only=lat_lon_only)
+        if out_path is not None:
+            hybrid_stats.save_stats(out_stats, out_path)
+
+        return out_stats
```
grantbuster, line 1588: Maybe split the file up? Getting big.

reVX/hybrid_stats/hybrid_stats.py
```diff
@@ -0,0 +1,1588 @@
...
+        Groupby data by month and/or hour
+
+        Parameters
+        ----------
+        data : pandas.DataFrame
+            DataFrame of data where index is time_index, columns are sites
+        diurnal : bool, optional
+            Extract diurnal stats, by default False
+        doy : bool, optional
+            Extract doy-of-year stats, by default False
```
grantbuster, line 340: If you're doing day-of-year statistics, should you add an option to roll the resource data to a local timezone? Otherwise the day of year is weird, right?

grantbuster (approved):

MRossol:
closes #83

PR #90, opened 2021-02-17
Distance to Ports

MRossol:


MRossol:
@grantbuster, this feature is "complete" and ready for review if you have time. A high level review is fine, but I would love your thoughts, inputs on the two similar run kwargs:
- cost_layer: this is the existing layer in the excl_h5 file that is used to create the cost layer, it needs to be the distance from each offshore pixel to land. In hindsight, I don't love the name, its ambiguous,  dist_layer might be better BUT:
- dist_layer: this is the layer to either save the distance_to_ports layer to, or to update (if update = True), maybe this should be ports_layer instead?

reVX/config/dist_to_ports.py
```diff
@@ -0,0 +1,71 @@
...
+
+    NAME = 'meanWindDirs'
+    REQUIREMENTS = ('ports_fpath', 'excl_fpath')
+
+    def __init__(self, config):
+        """
+        Parameters
+        ----------
+        config : dict
+            Dictionary with pre-extracted config input group.
```
grantbuster, line 20: Can't this also be a filepath to config?

reVX/offshore/dist_to_ports.py
```diff
@@ -0,0 +1,395 @@
...
+from reVX.wind_dirs import row_col_indices
+from rex.utilities.execution import SpawnProcessPool
+from rex.utilities.utilities import check_res_file
+
+logger = logging.getLogger(__name__)
+
+
+class DistanceToPorts:
+    """
+    Compute the distance to port in km
```
grantbuster, line 26: Maybe consider adding more description of the use case in the class docstring. For example, "use this class to compute the least cost distance from the exclusion grid pixels to port locations. The distance to shore layer will be used to calculate least cost paths around land masses and other such obstructions" or something like that?

And also maybe write down somewhere that you can update the least cost exclusion layer by adding in new ports? So like if you get one new port you can just run with a single input port and an existing least cost distnace layer? I think that is really powerful but kind of implicit right now.

reVX/offshore/dist_to_ports.py
```diff
@@ -0,0 +1,395 @@
...
+
+        port_coords = ports[['LATITUDE', 'LONGITUDE']].values
+        _, idx = tree.query(port_coords)
+
+        pixels = offshore_lat_lon.iloc[idx]
+
+        ports['row'] = pixels['row'].values
+        ports['col'] = pixels['col'].values
+        ports['dist_to_pixel'] = cls._haversine_distance(port_coords,
+                                                         pixel_coords[idx])
```
grantbuster, line 239: Is this mapping the ports to their respective nearest _single_ 90m pixels? If so, should specify in docstring. Current statement of "port locations and their mapping to the offshore pixels" makes it sound like it could be port -> 1 or more offshore pixels.

reVX/offshore/dist_to_ports.py
```diff
@@ -0,0 +1,395 @@
...
+            be the file into which the least cost distance to port is saved.
+        dist_layer : str, optional
+            Exclusions layer with distance to coast values,
+            by default 'dist_to_coast'
+        ports_layer : str, optional
+            Exclusion layer under which the distance to ports layer should be
+            saved, if None use the ports file-name, by default None
+        chunks : tuple, optional
+            Chunk size of dataset in .h5 file, by default (128, 128)
+        max_workers : int, optional
```
grantbuster, line 361: The parallelization scheme for this is by port right? So if adding one new port to the layer, max_workers should be 1?

reVX/plexos/plexos_plants.py
```diff
@@ -354,7 +354,7 @@ def _haversine_dist(cls, plant_coords, sc_coords):
         if plant_coords.shape[0] == 1:
             dist = dist.flatten()
 
-        R = 6373.0  # radius of the earth in kilometers
+        R = 6371.0  # radius of the earth in kilometers
```
grantbuster, line 357: Did the earth shrink?

tests/test_dist_to_ports.py
```diff
@@ -0,0 +1,119 @@
...
+    baseline = get_dist_to_ports(EXCL_H5)
+    dist = DistanceToPorts(PORTS_FPATH, EXCL_H5)
+    test = dist.least_cost_distance(max_workers=max_workers)
+
+    msg = 'distance to ports does not match baseline distances'
+    assert np.allclose(baseline, test), msg
+
+
+@pytest.mark.parametrize('ports_layer', [None, 'test'])
+def test_cli(runner, ports_layer):
```
grantbuster, line 57: I thought you were going to keep the heatmap plot output as an option in the test for this feature? This is definitely a feature thats easier to visualize.

grantbuster (changes requested):

grantbuster:
@MRossol what about `input_dist_layer` and `output_dist_layer`?

reVX/config/dist_to_ports.py
MRossol: good question, do you know? ;-) I can look

reVX/offshore/dist_to_ports.py
```diff
@@ -0,0 +1,395 @@
...
+class DistanceToPorts:
+    """
+    Compute the distance to port in km
+    """
+    def __init__(self, ports, excl_fpath, layer='dist_to_coast'):
+        """
+        Parameters
+        ----------
+        ports : str
+            Path to shape file containing ports to compute least cost distance
```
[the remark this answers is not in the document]
MRossol: That's what I got from the GDS team, technically your correct. I'll make the parser more dynamic

reVX/offshore/dist_to_ports.py
```diff
@@ -0,0 +1,395 @@
...
+        ports : geopandas.GeoDataFrame
+            DataFrame of port locations and their mapping to the offshore
+            pixels for least cost distance computation
+        """
+        ports = gpd.read_file(ports, ignore_geometry=True)
+
+        pixel_coords = offshore_lat_lon[['latitude', 'longitude']].values
+        tree = cKDTree(pixel_coords)  # pylint: disable=not-callable
+
+        port_coords = ports[['LATITUDE', 'LONGITUDE']].values
```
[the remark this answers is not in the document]
MRossol: Good catch! Yes it should, that is a super handy new util!

reVX/config/dist_to_ports.py
grantbuster: haha yes, str filepath to json config or pre-extracted config in dict form

reVX/offshore/dist_to_ports.py
```diff
@@ -0,0 +1,395 @@
...
+            by default 'dist_to_coast'
+
+        Returns
+        -------
+        arr : ndarray
+            Cost array with offshore pixels set to 90 (pixel width) and on
+            shore pixels set to 9999.
+        profile : dict
+            Profile (transform, crs, etc.) of arr raster
+        lat_lon : pandas.DataFrame
```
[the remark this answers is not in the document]
MRossol: this is all based on the exclusion arrays. This dataframe is a flattened list of the lat, lon, row, col coordinates for all offshore pixels. I'll make the docstrings clearer, you've really raised our docstring bar, I think its great but man are our old modules going to look bad ;-)

reVX/offshore/dist_to_ports.py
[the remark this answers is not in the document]
grantbuster: Yeah Im not surprised that they would have shape files inputs but it seems like all we need is point locations so like a meta data file might be worth handling.

reVX/offshore/dist_to_ports.py
MRossol: not quite, it is to the nearest offshore 90m pixel. The issue is that sometimes the ports sit inland based on our dist_to_coast raster

reVX/offshore/dist_to_ports.py
MRossol: yes, but if your run with update_layer it could have multiple new ports in the shape file (or future .csv)

reVX/plexos/plexos_plants.py
MRossol: I for the life of me could not figure out where you got 6373! Everything i found online said 6371, we should get consensus on that...

tests/test_dist_to_ports.py
MRossol: brain fart, I'll add in that function.

MRossol:
input and output dist_layer works for me!

reVX/offshore/dist_to_ports.py
MRossol: Can do

reVX/offshore/dist_to_ports.py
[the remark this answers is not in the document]
grantbuster: Haha yes most of the bar raising is because i cant figure out how to use our old code....

reVX/plexos/plexos_plants.py
grantbuster: Fortunately i think this is just a scalar on the distance so the relative and nearest distances should be all good. The radius depends on where you're measuring since the earth is not a perfect sphere. seems to be 6378 at the equator. But yes i'm also seeing 6371 as the average. Might have been a typo....

MRossol:
@grantbuster  I think I got all of the changes

grantbuster (approved):

PR #93, opened 2021-03-02
Gb/rrp shapes

grantbuster:


grantbuster:
Here's the plot produced by the rev-reeds-plexos test with the shape file definition of the "plexos nodes". Each shape will aggregate all resident supply curve points with respective capacities requested by reeds. This example results in 2 output power profiles. 
test_rev_reeds_plexos_shape

MRossol:
Cool!

reVX/plexos/rev_reeds_plexos.py
```diff
@@ -4,19 +4,22 @@
...
 import pandas as pd
 from scipy.spatial import cKDTree
+from geopandas import GeoDataFrame
 from warnings import warn
 
 from rex.rechunk_h5 import to_records_array
 from rex.utilities.execution import SpawnProcessPool
 from rex.utilities.utilities import parse_table
 
+from reVX.utilities.region import RegionClassifier
```
MRossol, line 22: Silly i know but I think we should rename the module region_classifier

reVX/plexos/rev_reeds_plexos.py
```diff
@@ -38,9 +48,13 @@ def __init__(self, sc_build, cf_fpath, cf_res_gids=None,
...
         cf_fpath : str
             File path to capacity factor file to get profiles from.
-        cf_res_gids : list | np.ndarray, optional
+        res_gids : list | np.ndarray, optional
             Resource GID's available in cf_fpath, if None pull from cf_fpath,
             by default None
+        force_full_build : bool
+            Flag to ensure the full requested buildout is built at each SC
+            point. If True, the remainder of the requested build will always
+            be built at the last resource gid in the sc point.
```
MRossol, line 57: Do we have a check for the 1e6 hack? I.e. what if the requested build is > the capacity of the last resource gid? Do we just multiply the cf of the last resource gid by the requested capacity?

reVX/plexos/rev_reeds_plexos.py
```diff
@@ -465,9 +571,25 @@ def _parse_plexos_nodes(plexos_nodes):
         plexos_nodes : pd.DataFrame
             Plexos node meta data including gid, lat/lon, voltage
         """
-        plexos_nodes = parse_table(plexos_nodes)
-        plexos_nodes = DataCleaner.pre_filter_plexos_meta(plexos_nodes)
+
+        if (isinstance(plexos_nodes, str)
```
MRossol, line 575: Why not add a sub if/else under isisnstance(plexos_nodes, str)?

reVX/plexos/rev_reeds_plexos.py
```diff
@@ -465,9 +571,25 @@ def _parse_plexos_nodes(plexos_nodes):
...
             Plexos node meta data including gid, lat/lon, voltage
         """
-        plexos_nodes = parse_table(plexos_nodes)
-        plexos_nodes = DataCleaner.pre_filter_plexos_meta(plexos_nodes)
+
+        if (isinstance(plexos_nodes, str)
+                and plexos_nodes.endswith(('.csv', '.json'))):
+            plexos_nodes = parse_table(plexos_nodes)
+        elif isinstance(plexos_nodes, str) and plexos_nodes.endswith('.shp'):
+            rc = RegionClassifier(self.sc_build, plexos_nodes,
```
MRossol, line 579: do we want to un-hide RegionsClassifier._regions?

reVX/plexos/rev_reeds_plexos.py
```diff
@@ -465,9 +571,25 @@ def _parse_plexos_nodes(plexos_nodes):
...
+                                  regions_label=None)
+            plexos_nodes = rc._regions
+            if 'plexos_id' not in plexos_nodes:
+                plexos_nodes['plexos_id'] = np.arange(len(plexos_nodes))
+
+        elif not isinstance(plexos_nodes, pd.DataFrame):
+            msg = ('Did not recognize plexos_nodes input: {}'
```
MRossol, line 586: We should make the error message more specific, i.e. expecting x, y, or z, but got:

MRossol (commented):
Are you still planning on making a basic plant aggregator or do you want me to do that as a parent class for PlexosPlants? I feel like it would be useful, but its still unclear how applicable it is for this workflow.

reVX/plexos/rev_reeds_plexos.py
grantbuster: like the file? yes, agreed!

reVX/plexos/rev_reeds_plexos.py
grantbuster: Yeah this flag turns on the 1e6 hack and i added a negative test with the flag=false to ensure that it will fail if the requested build > available capacity within one watt. If this flag is true, the last GID in the supply curve point is given a 1e6MW capacity ("the remainder of the requested build will always be built at the last resource gid in the sc point")

reVX/plexos/rev_reeds_plexos.py
grantbuster: same effect isnt it? And adds "complexity" in the form of more levels. Bad incentives by pylint!!

reVX/plexos/rev_reeds_plexos.py
grantbuster: Yep can do!

PR #96, opened 2021-03-10
Offshore inputs

MRossol:


MRossol:
@grantbuster, FYI I need to add in the assembly area nonsense, will do that asap

reVX/config/offshore_inputs.py
```diff
@@ -0,0 +1,58 @@
...
+        (required)
+        """
+        return self['offshore_sites']
+
+    @property
+    def input_layers(self):
+        """
+        Path to json file containing input layer, list of input layers, to
+        extract, or dictionary mapping the input layers to extract to
+        the column names to save them under (required)
```
grantbuster, line 48: which is the key, which is the value?

reVX/offshore/offshore_inputs.py
```diff
@@ -0,0 +1,300 @@
...
+from reV.handlers.exclusions import ExclusionLayers
+
+logger = logging.getLogger(__name__)
+
+
+class OffshoreInputs:
+    """
+    Class to extract offshore inputs from offshore inputs .h5
+    """
+    def __init__(self, inputs_fpath, offshore_sites, tm_dset='techmap_wtk'):
```
grantbuster, line 22: docstring?

reVX/offshore/offshore_inputs.py
```diff
@@ -0,0 +1,300 @@
...
+                   .format(list(offshore_sites.columns)))
+            logger.error(msg)
+            raise RuntimeError(msg)
+
+        offshore_gids = offshore_sites['gid'].values
+        tech_map = self._reduce_tech_map(self._inputs_fpath, tm_dset=tm_dset,
+                                         offshore_gids=offshore_gids)
+
+        offshore_meta = pd.merge(offshore_sites, tech_map, on='gid',
+                                 how='inner')
```
grantbuster, line 235: inner seems dangerous here. Could we unknowingly drop offshore sites? Maybe a simple assert checking that the offshore_meta output has the same length as the offshore_sites input?

reVX/offshore/offshore_inputs.py
```diff
@@ -0,0 +1,300 @@
...
+
+        Returns
+        -------
+        offshore_sites : pandas.DataFrame
+            Offshore sites meta data
+        """
+        if isinstance(offshore_sites, str):
+            if offshore_sites.endswith('.h5'):
+                with Resource(offshore_sites) as f:
+                    offshore_sites = f.meta.reset_index()
```
grantbuster, line 184: Is this how you're getting "gid" as a column? Kind of implicit, hard to read, subject to index naming conventions of the Resource handler.

reVX/offshore/offshore_inputs.py
```diff
@@ -0,0 +1,300 @@
...
+        Returns
+        -------
+        out : pandas.DataFrame
+            Updated meta data table with desired layers
+        """
+        if isinstance(layers, str):
+            layers = [layers]
+
+        if isinstance(layers, (tuple, list, np.ndarray)):
+            layers = {layer: layer for layer in layers}
```
grantbuster, line 52: lol, python

tests/test_offshore_inputs.py
```diff
@@ -0,0 +1,103 @@
...
+        msg = ('Failed with error {}'
+               .format(traceback.print_exception(*result.exc_info)))
+        assert result.exit_code == 0, msg
+
+        test = os.path.basename(input_layers).replace('.json', '.csv')
+        test = os.path.join(td, test)
+        test = pd.read_csv(test)
+
+        baseline = pd.read_csv(BASELINE)
+        assert_frame_equal(baseline, test, check_dtype=False)
```
grantbuster, line 81: I think real checks would be valuable here. For example: check the offshore meta data and go manually retrieve the row/col indices from the h5 and assert equals. Check that the offshore gid to exclusion h5 index is done correctly.

grantbuster (changes requested):
Very simple, nice work.

reVX/offshore/offshore_inputs.py
MRossol: This was sort of a result of our test data. It turns out that the offshore exclusions don't reach as far out as the resource data, so some of the offshore sites weren't in the techmap. I added this warning:
https://github.com/NREL/reVX/blob/5ac8db1c09519d29b835002f90a3f949420f12ca/reVX/offshore/offshore_inputs.py#L145-L152

reVX/offshore/offshore_inputs.py
MRossol: I'll make it more explicit

reVX/offshore/offshore_inputs.py
MRossol: seriously!

tests/test_offshore_inputs.py
MRossol: Good call

MRossol:
@grantbuster,

I think we are all good minus array efficiency. Take a gander when you have a chance if you don't mind.

One question, should we merge this and add array efficiency as a new feature, or leave this open and wait? My preference would be the former just because I don't like stale branches...

reVX/offshore/dist_to_ports.py
```diff
@@ -397,21 +431,227 @@ def run(cls, ports, excl_fpath, input_dist_layer='dist_to_coast',
...
                                                 max_workers=max_workers)
-        dtp.save_as_layer(output_dist_layer, dist_to_ports, chunks=chunks)
+
+        if output_dist_layer:
+            dtp.save_as_layer(output_dist_layer, dist_to_ports, chunks=chunks)
 
         return dist_to_ports
+
+
+class AssemblyAreas:
```
grantbuster, line 454: I think this class and its CLI should be in their own files. That would make it much easier to see the available utilities at a glance. otherwise you need to run the cli, scroll through this file, or know exactly what you're looking for.

reVX/offshore/dist_to_ports.py
```diff
@@ -397,21 +431,227 @@ def run(cls, ports, excl_fpath, input_dist_layer='dist_to_coast',
...
+            Shape of exclusion layers/arrays
+        """
+        with ExclusionLayers(excl_fpath) as f:
+            lat = f['latitude']
+            lon = f['longitude']
+
+        lat_lon = np.dstack((lat.ravel(), lon.ravel()))[0]
+
+        # pylint: disable=not-callable
+        tree = cKDTree(lat_lon)
```
grantbuster, line 517: Is this slow? Could we only make a tree on the offshore exclusion points?

grantbuster (changes requested):
Looks wicked simple. A few minor recommended changes.

reVX/offshore/dist_to_ports.py
MRossol: Fixed, not to bad, great minds!

MRossol:
@grantbuster, I moved assembly areas to its own module and CLI, BUT i did leave the option in the dist_to_ports cli to run assembly areas at the end. thoughts?

grantbuster:
looks great! I think thats a nice feature.

grantbuster (approved):

PR #99, opened 2021-03-24
added share_resource flag to simple plant builder with check in self.…

grantbuster:
…check_valid_buildouts() and pytest

### DIFF
--- a/reVX/plexos/simple_plant_builder.py
+++ b/reVX/plexos/simple_plant_builder.py
@@ -255,7 +255,12 @@ class SimplePlantBuilder(BaseProfileAggregation):
         for i, single_plant_sc in plant_sc_builds.items():
             sc_res_gids = single_plant_sc['res_gids'].values.tolist()
             sc_res_gids = [g for subset in sc_res_gids for g in subset]
-            missing = [gid for gid in sc_res_gids
+            gid_caps = single_plant_sc['gid_capacity'].values.tolist()
+            gid_caps = [g for subset in gid_caps for g in subset]
+            assert len(gid_caps) == len(sc_res_gids)
+            plant_built_res_gids = [gid for j, gid in enumerate(sc_res_gids)
+                                    if gid_caps[j] > 0]
+            missing = [gid for gid in plant_built_res_gids
                        if gid not in self.available_res_gids]
             if any(missing):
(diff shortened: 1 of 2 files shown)


reVX/plexos/simple_plant_builder.py
```diff
@@ -214,12 +218,15 @@ def assign_plant_buildouts(self):
                         gids_remain[j] -= built
                         plant_cap_to_build -= built
 
+                        # if specified to not share resource, set remaining
+                        # capacity for this resource gid to zero
+                        if not self._share_res:
```
MRossol, line 223: so simple!

reVX/plexos/simple_plant_builder.py
```diff
@@ -254,6 +263,24 @@ def check_valid_buildouts(self, plant_sc_builds):
...
+            gid_caps = [g for subset in gid_caps for g in subset]
+            assert len(gid_caps) == len(sc_res_gids)
+
+            plant_res_gids = [gid for i, gid in enumerate(sc_res_gids)
+                              if gid_caps[i] > 0]
+
+            shared = [gid for gid in plant_res_gids
+                      if gid in global_built_res_gids]
+
+            if any(shared) and not self._share_res:
```
MRossol, line 276: Can this happen? Do we have a negative test for this? Or is this just a triple check?

MRossol (commented):
One question, but the meat looks nice and simple!

reVX/plexos/simple_plant_builder.py
grantbuster: I built in this check first as more of full-scale on-the-fly check as opposed to the pytests. This method is useful to raise any errors before we start running parallel compute. But no this should never happen, I built it in first and made sure that it would raise a good error then built the feature. It's hard to do a negative test because if you tell the object to not share resource it performs property and won't raise an error. I did add a test to make sure that it won't share resource though.

MRossol: Gotcha, I'm honestly torn, its a good check but if its never going to run I wonder if its needed? Honestly partly thinking of our code coverage score ;-), but also its fairly complex and if its not needed I'm okay removing it and relying on the pytest and your super simple elegent solution!

MRossol (approved):
once tests pass I approve

PR #100, opened 2021-03-24
better logic on simple plant builder to always build by iterating thr…

grantbuster:
…ough resource gids within sc points and only build if not shared


MRossol (approved):
looks good on the surface

Issue #70, opened 2020-12-08
Shadow flicker wind exclusions 


grantbuster:
**Why this feature is necessary:**
Wind build regulations are limiting the amount of shadow flicker (hours per year) on households/residences/buildings. We have a shadow flicker model (from Darice) and need to think about how to use this to create an exclusions file for ingestion into reV. 

**A possible solution is:**
From discussion on 12/8/2020 we think the steps required are:
1. get 90m building/residences raster
2. run shadow flicker for ever SC point for one turbine technology (what kind of wind resource do we need? Wind speed AND direction? Just solar position?)
3. make a database of high res shadow flicker shapes dependent on geospatial info (solar and wind)
4. map high res shadow flicker database to 90m resolution
5. map 90m res shadown flicker to an exclusions file (iterate through buildings and "back out the exclusions" from that shadow)

**Charge code**
WEPG.10348.08.01.01

**Urgency / Timeframe**
?

MRossol:
Inputs:
- lat
- lon
- blade_length (m)
- wind_dir (optional), degrees from N

Examples: 200 x 300m heat map

grantbuster:
Here's a bad drawing of the desired exclusion mask. The turbine flicker shape should be computed on a 270x270m output array, which will be converted to a 3x3 array of annual flicker fraction or a boolean, and then we need to iterate through 90m pixels and exclude those that would shade/flicker a building. 

![image](https://user-images.githubusercontent.com/12700356/103557519-9222f400-4e70-11eb-8062-8b2f136d76b7.png)

MRossol:
Plan:
- Run HOPP to produce a 270m x 270m heat map (3x3 90m array). Convert shadow profile into a boolean mask
- Compute i,j coordinates that would cause exclusion.
- Check if a building is in the above i,j vector

PR #109, opened 2021-04-29
Shadow flicker

MRossol:
Turbine Flicker Exclusions
Bug fixes


MRossol:
@grantbuster, the tests pass locally, but HOPP is breaking the actions :-(

MRossol:
closes #70

reVX/turbine_flicker/turbine_flicker.py
```diff
@@ -0,0 +1,563 @@
+# -*- coding: utf-8 -*-
+"""
+Turbine Flicker exclusions calculator
+"""
+from concurrent.futures import as_completed
+from hybrid.flicker.flicker_mismatch_grid import FlickerMismatch
```
grantbuster, line 6: Do we want to have this as an import in the init statement and not in the requirements? How heavy of an install is the hybrid library?

reVX/turbine_flicker/turbine_flicker.py
```diff
@@ -0,0 +1,563 @@
...
+        max_workers : None | int, optional
+            Number of workers to use, if 1 run in serial, if None use all
+            available cores, by default None
+        out_layer : str, optional
+            Layer to save exclusions under. Layer will be saved in
+            "excl_fpath", by default None
+
+        Returns
+        -------
+        flicker_excl : ndarray
```
grantbuster, line 549: Is this an exclusion array where 1 is excluded or an inclusion array where 1 is included? Applicable to all instances of this docstring.

reVX/turbine_flicker/turbine_flicker.py
```diff
@@ -0,0 +1,563 @@
...
+            sc_blds = f[building_layer, row_slice, col_slice]
+
+        row_idx = np.array(range(*row_slice.indices(row_slice.stop)))
+        col_idx = np.array(range(*col_slice.indices(col_slice.stop)))
+        bld_row_idx, bld_col_idx = np.where(sc_blds > building_threshold)
+
+        return row_idx[bld_row_idx], col_idx[bld_col_idx], shape
+
+    @classmethod
+    def _exclude_turbine_flicker(cls, gid, excl_fpath, res_fpath,
```
grantbuster, line 294: Does this method apply the shadow-based exclusion to _all possible_ exclusion pixels using the buildings in the specified SC point gid (including excluding pixels outside of the specified SC point gid)? Or does it only look at buildings within and resulting exclusions within the specified SC point?

reVX/turbine_flicker/turbine_flicker.py
```diff
@@ -0,0 +1,563 @@
...
+                                                         ("time", ))[0]
+
+        return shadow_flicker
+
+    @staticmethod
+    def _check_shadow_flicker_arr(shadow_flicker):
+        """
+        Check to ensure the shadow_flicker array is odd in shape, i.e. both
+        dimensions are odd allowing for a central pixel for the turbine to
+        sit on. Flip 0-axis to mimic the turbine sitting on each building.
```
grantbuster, line 165: Is that really all you have to do to back out the excluded turbine location from a building using the shadow profile? Why wouldn't you have to also flip the 1-axis?

reVX/turbine_flicker/turbine_flicker.py
```diff
@@ -0,0 +1,563 @@
...
+
+        Returns
+        -------
+        row_shifts : ndarray
+            Shifts along axis 0 from building location to pixels to be excluded
+        col_shifts : ndarray
+            Shifts along axis 1 from building location to pixels to be excluded
+        """
+        # ensure shadow_flicker array is regularly shaped
+        shadow_flicker = cls._check_shadow_flicker_arr(shadow_flicker)
```
grantbuster, line 224: Maybe update the variable + method name because isnt the output now an inverted mapping of exclusion shifts? This is kind of black magic and hard to think about without seeing but if you've checked the output visually then i trust ya and its a great elegant solution!

reVX/turbine_flicker/turbine_flicker.py
```diff
@@ -0,0 +1,563 @@
...
+        Parameters
+        ----------
+        shadow_flicker : ndarray
+            2D array centered on the turbine with the number of flicker hours
+            per "exclusion" pixel
+
+        Returns
+        -------
+        shadow_flicker : ndarray
+            Updated 2D shadow flicker array with odd dimensions if needed
```
grantbuster, line 178: And also inverted right? seems important that someone should not expect basically the same outputs as input.

reVX/turbine_flicker/turbine_flicker.py
```diff
@@ -0,0 +1,563 @@
...
+        blade_length = hub_height / 2.5
+        shadow_flicker = cls._compute_shadow_flicker(meta['latitude'],
+                                                     meta['longitude'],
+                                                     blade_length,
+                                                     wind_dir)
+
+        row_shifts, col_shifts = cls._threshold_flicker(
+            shadow_flicker, flicker_threshold=flicker_threshold)
+
+        excl_row_idx = (row_idx + row_shifts[:, None]).ravel()
```
grantbuster, line 360: Once again this is really hard to think about without just seeing but if you've checked the outputs visually this is great!!

grantbuster (changes requested):

reVX/turbine_flicker/turbine_flicker.py
MRossol: That is a very good question. From PIP its not to bad. Its not an easy install from Conda. I'm fine with making it be a conditional requirement just for Turbine/shadow flicker

reVX/turbine_flicker/turbine_flicker.py
MRossol: Very good catch, we should really re-name everything to be inclusion mask. will change

reVX/turbine_flicker/turbine_flicker.py
MRossol: All of these methods are of a single SC point gid. It will only map based on buildings within the SC Point gid sub-section of the exclusions array, BUT it will exclude pixels outside of the sub-section, if for example a building is right no the edge.

grantbuster:
@MRossol also one more note: I would recommend trying to set up a unit test where you give it a single building and a flicker profile with one shadow-d pixel (maybe +2, +2) and make sure that it excludes the correct pixel (-2, -2). This could actually be parameterized and you could run a few different super simple shadow profiles. This would just be a really good sanity check.

reVX/turbine_flicker/turbine_flicker.py
MRossol: hmm, good question about the 1-axis flip. I'll have to mock that up.

reVX/turbine_flicker/turbine_flicker.py
MRossol: I'll add the plots to the tests, but yeah for matrix math!

reVX/turbine_flicker/turbine_flicker.py
grantbuster: I dont have a strong opinion and will definitely defer to you. It would definitely be nice to be able to run flicker tests with actions. But if the hybrid install isn't something we're going to use much it might be nice as an optional requirement. Once again, totally going to defer to you.

reVX/turbine_flicker/turbine_flicker.py
grantbuster: Inclusion masks are way better + more useful anyway. I hate the naming issues around this.

MRossol:
@grantbuster, I tried to do that here:
https://github.com/NREL/reVX/blob/f9c5baa2e9284e4854198d22d17f2d0ad00740ed/tests/test_turbine_flicker.py#L37

I'll tweak the code to create a function that takes the shadow flicker array and the building array and return the shifts to make a unit test easier

reVX/turbine_flicker/turbine_flicker.py
grantbuster: Okay fantastic. That's exactly what I was getting at. Great. Maybe if you feel like it, add that note to the docstring that it will exclude pixels outside the sub-array

reVX/turbine_flicker/turbine_flicker.py
grantbuster: I think we have to flip on both axis, see my comment about a unit test below :)

MRossol: I think you might be right... Conceptually this problem makes my brain hurt....

grantbuster:
Yeah gotcha but I really think we need to flip along both axes because as-is i think a shadow of (+2, +2) will exclude (-2, +2) but should be (-2, -2). It would be nice to be able to test this intuitively and not from a baseline test.

MRossol:
I agree, I'll make the changes and put in a cleaner unit-test with synthetic building and shadow flicker data

MRossol:
@grantbuster  is this more of what you were thinking:
https://github.com/NREL/reVX/blob/6ff8be9620baddadcceb50ba3becdb1013b7ed5f/tests/test_turbine_flicker.py#L61-L81

reVX/turbine_flicker/turbine_flicker.py
grantbuster: still one place you should fix: https://github.com/NREL/reVX/blob/c4fe5f472c63effd0792153b54195780985372aa/reVX/turbine_flicker/turbine_flicker.py#L592

reVX/turbine_flicker/turbine_flicker.py
```diff
@@ -0,0 +1,605 @@
...
+                                         gridcell_height=cls.GRIDCELL_SIZE,
+                                         gridcell_width=cls.GRIDCELL_SIZE,
+                                         gridcells_per_string=1)
+        shadow_flicker = shadow_flicker.create_heat_maps(range(0, 8760),
+                                                         ("time", ))[0]
+
+        return shadow_flicker
+
+    @staticmethod
+    def _check_shadow_flicker_arr(shadow_flicker):
```
grantbuster, line 163: I still think you should rename this method since it's doing more than just checking the shape - it is inverting the whole array.

grantbuster:
For tests that is more of what i was thinking but honestly I'm thinking even more simply. Your test still assumes that the 1st and 2nd axis inversion is the correct method. What I'm thinking is something that would check this incredibly simple human-interpretable test where a single shadow pixel and a single building can be easily understood:

image

For example, pseudocode of the parameterization:

```
@pytest.mark.parametrize(('shadow_vector', 'excl_vector'), ((-2, -2), (2, 2)))
def test_simple(shadow_vector, excl_vector):
shadow_arr  = zeros()
shadow_arr[shadow_vector] = 1
incl = ShadowFlicker.exclude_shadow(shadow_arr)
assert incl[excl_vector] == 0
incl[excl_vector] == 1
assert (incl == 1).all()

reVX/turbine_flicker/turbine_flicker.py
MRossol: doh

reVX/turbine_flicker/turbine_flicker.py
MRossol: LOL, I think I renamed the wrong method, oh well baby steps.

MRossol:
I gotcha now. I'll add that, more tests the better right. It will be less straight forward since the vectors don't map directly to positional indices.

grantbuster:
Definitely the more tests the better! I feel strongly that this kind of a stupid simple sanity check test will be more useful and also easier to maintain than a test that saves a baseline (that sucks to re-verify) or a test that implicitly assumes any part of the methodology is correct.

MRossol:
Agreed, I've been trying really hard to add in both. Baseline for production, unit test for sanity/clarity

grantbuster (approved):
Yep this is exactly what i had in mind. Nice!

PR #112, opened 2021-05-13
Use nearest resource wind direction time-series instead of aggregate mean

MRossol:


grantbuster (approved):
I'm way too separated from this work to really comment intelligently without spending significant time learning the code

MRossol:
LOL, no worries more just wanted an extra set of eye to make sure I didn't do anything obviously stupid or heinous

PR #121, opened 2021-07-20
Xmission/costs handler

MRossol:
Add transmission cost handler class based on ExtractionLayers

### DIFF
--- a/reVX/handlers/geotiff.py
+++ b/reVX/handlers/geotiff.py
@@ -64,14 +64,16 @@ class Geotiff:
             Slicing args similar to a numpy array slice. See examples above.
         """
         ds, ds_slice = parse_keys(keys)
+        out = None
+        if isinstance(ds, str):
+            if ds == 'meta':
+                out = self._get_meta(*ds_slice)
+            elif ds.lower().startswith('lat'):
+                out = self._get_lat_lon(*ds_slice)[0]
+            elif ds.lower().startswith('lon'):
+                out = self._get_lat_lon(*ds_slice)[1]
 
-        if ds == 'meta':
-            out = self._get_meta(*ds_slice)
-        elif ds.lower().startswith('lat'):
-            out = self._get_lat_lon(*ds_slice)[0]
-        elif ds.lower().startswith('lon'):
-            out = self._get_lat_lon(*ds_slice)[1]
-        else:
+        if out is None:
             out = self._get_data(ds, *ds_slice)
 
         return out
@@ -374,6 +376,7 @@ class Geotiff:
         """
         y_slice, x_slice = self._unpack_slices(*ds_slice)
         data = self._src.data[ds, y_slice, x_slice].flatten().compute()
+
         return data
 
     def _create_profile(self, chunks=(128, 128)):
--- a/reVX/handlers/xmission_cost_layers.py
+++ b/reVX/handlers/xmission_cost_layers.py
@@ -64,8 +64,9 @@ class XmissionCostsLayers(ExclusionLayers):
         layers = ['base_costs', 'multipliers_{}mw'.format(self._capacity)]
         for layer_name in layers:
             if layer_name not in self.layers:
-                msg = ('{} not in available layers: {}'
-                       .format(layer_name, self.layers))
+                msg = ('{} is needed to compute costs but is not in available '
+                       'layers: {}. Please add it to {} and try again!'
+                       .format(layer_name, self.layers, self.h5_file))
                 logger.error(msg)
                 raise HandlerKeyError(msg)
 
@@ -91,8 +92,10 @@ class XmissionCostsLayers(ExclusionLayers):
                   'transmission_barrier']
         for layer_name in layers:
             if layer_name not in self.layers:
-                msg = ('{} not in available layers: {}'
-                       .format(layer_name, self.layers))
+                msg = ('{} is needed to compute mcp_costs but is not in '
+                       'available layers: {}. Please add it to {} and try '
+                       'again!'
+                       .format(layer_name, self.layers, self.h5_file))
                 logger.error(msg)
                 raise HandlerKeyError(msg)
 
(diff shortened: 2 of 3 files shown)


reVX/handlers/xmission_cost_layers.py
```diff
@@ -13,3 +15,89 @@ class XmissionCostsLayers(ExclusionLayers):
...
+        capacity : int
+            Capacity bin to extract costs for
+        hsds : bool, optional
+            Boolean flag to use h5pyd to handle .h5 'files' hosted on AWS
+            behind HSDS, by default None
+        """
+        self._capacity = capacity
+        super().__init__(h5_file, hsds=hsds)
+
+    def __getitem__(self, keys):
```
mikebannis, line 34: Why not use `@property` for these items instead of `__getitem__`?

reVX/handlers/xmission_cost_layers.py
```diff
@@ -13,3 +15,89 @@ class XmissionCostsLayers(ExclusionLayers):
...
+        ds, ds_slice = parse_keys(keys)
+
+        if ds.lower().startswith('lat'):
```
mikebannis, line 37: Understood you're trying to catch lat and latitude, but it also allows misspellings and other unexpected options to successfully return data.

reVX/handlers/xmission_cost_layers.py
```diff
@@ -13,3 +15,89 @@ class XmissionCostsLayers(ExclusionLayers):
...
+            tuple describing slice of cost array to extract
+
+        Returns
+        -------
+        costs : ndarray
+            Array of raw tie-line costs
+        """
+        layers = ['base_costs', 'multipliers_{}mw'.format(self._capacity)]
+        for layer_name in layers:
+            if layer_name not in self.layers:
```
mikebannis, line 66: Why would a layer not exist? Not created yet?

reVX/handlers/xmission_cost_layers.py
```diff
@@ -13,3 +15,89 @@ class XmissionCostsLayers(ExclusionLayers):
...
+
+        Returns
+        -------
+        mcp_costs : ndarray
+            Array of MinimumCostPath costs
+        """
+        layers = ['base_costs', 'multipliers_{}mw'.format(self._capacity),
+                  'transmission_barrier']
+        for layer_name in layers:
+            if layer_name not in self.layers:
```
mikebannis, line 93: Same, why would layer not exist? Error message should be more explicit.

reVX/handlers/xmission_cost_layers.py
```diff
@@ -13,3 +15,89 @@ class XmissionCostsLayers(ExclusionLayers):
...
+                msg = ('{} not in available layers: {}'
+                       .format(layer_name, self.layers))
+                logger.error(msg)
+                raise HandlerKeyError(msg)
+
+        mcp_costs = (self[(layers[0], ) + ds_slice]
```
mikebannis, line 99: This feels obfuscated. Can layers[1] etc be replaced with the appropriate string? Or use a constant so that the strings don't get defined in multiple places. That's a bit harder for 'multipliers_{}mw'.

reVX/handlers/xmission_cost_layers.py
MRossol: __getitem__ allows for flexibilty, its executed as self[...] and as you said we can allow for custom functions. here we are adding in the ability to "get" "costs" and "mcp_costs" by computing them on the fly

reVX/handlers/xmission_cost_layers.py
MRossol: Exactly, not created yet. My long term goal for all of this code is to make it available to the public and its a very quick simple check to just make sure the layer exists. we also raise a more verbose error than the basic key error you'd get form h5py by listing all available layers. I'll add more comments that the layers are needed to compute...

reVX/handlers/xmission_cost_layers.py
MRossol: I could repeat the strings but since the list is really short its pretty easy/quick to just go look at value 0, 1, and 2 to figure out the logic. I was trying to not write out the strings multiple times. As you pointed out it gets tedious with the multipliers layer

reVX/handlers/geotiff.py
```diff
@@ -64,14 +64,16 @@ def __getitem__(self, keys):
...
+        out = None
+        if isinstance(ds, str):
+            if ds == 'meta':
+                out = self._get_meta(*ds_slice)
+            elif ds.lower().startswith('lat'):
+                out = self._get_lat_lon(*ds_slice)[0]
+            elif ds.lower().startswith('lon'):
+                out = self._get_lat_lon(*ds_slice)[1]
+
+        if out is None:
```
mikebannis, line 76: How about just `else:` for line 76? Should this throw an error before returning `None`?

MRossol: So the issue you ran into before is that the if/elif/elif statement ASSUMES `ds` is a string, but the bands are normally numerical. The most verbose way of doing this would be:
```
        ds, ds_slice = parse_keys(keys)

        if isinstance(ds, str):
            if ds == 'meta':
                out = self._get_meta(*ds_slice)
            elif ds.lower().startswith('lat'):
                out = self._get_lat_lon(*ds_slice)[0]
            elif ds.lower().startswith('lon'):
                out = self._get_lat_lon(*ds_slice)[1]
           else:
                out = self._get_data(ds, *ds_slice)
       else:
            out = self._get_data(ds, *ds_slice)

        return out
```

But that seems silly redudant, instead I just check for meta, lat, or lon, otherwise just pull the given ds value

mikebannis: 👍

mikebannis (approved):

MRossol:
@mikebannis  do you want to merge this into your branch (FYI thats what the PR is setup to do)

PR #122, opened 2021-07-22
Xmission/cost layers

mikebannis:


reVX/least_cost_xmission/cost_creator.py
```diff
@@ -0,0 +1,304 @@
...
+        default_mults : dict
+            Default line cost multipliers for areas not covered by the ISO
+            regions.
+        """
+        logger.info('Loading all files')
+        with Geotiff(iso_regions_f) as gt:
+            self._iso_regions = gt.values[0]
+            self._profile = gt.profile
+
+        with ExclusionLayers(excl_h5) as el:
```
MRossol, line 59: Move these to the create_multipliers method, they aren't needed here

reVX/least_cost_xmission/cost_creator.py
```diff
@@ -0,0 +1,304 @@
...
+            # self._write_geotiff('out/land_use.tif', self._profile, self._land_use)
+
+        assert self._iso_regions.shape == self._land_use.shape == \
+            self._slope.shape, 'All rasters must be the same shape'
+        logger.info('Loading files complete')
+
+        self._profile['dtype'] = 'float32'
+        self._profile['nodata'] = None
+
+        self._land_use_classes = land_use_classes
```
MRossol, line 72: Where are these used, lets move them to that method to make the init a bit cleaner

reVX/least_cost_xmission/cost_creator.py
```diff
@@ -0,0 +1,304 @@
...
+        self._iso_mults = iso_mults
+        self._base_line_costs = base_line_costs
+        self._iso_lookup = iso_lookup
+        self._default_mults = default_mults
+        self._cell_size = CELL_SIZE
+
+    @classmethod
+    def build_cost_rasters(cls, iso_regions_f,
```
MRossol, line 80: My preference is to order args/kwargs in the order that they are used in the class method:
1) init
2) first method called
3)...

reVX/least_cost_xmission/cost_creator.py
```diff
@@ -0,0 +1,304 @@
...
+                            base_line_costs_f=base_line_costs_f,
+                            iso_lookup_f=iso_lookup_f,
+                            power_classes_f=power_classes_f)
+
+        xcc = cls(iso_regions_f, land_use_classes, xc.iso_mults,
+                  xc.base_line_costs, xc.iso_lookup, default_mults)
+        mults_arr = xcc.create_multipliers()
+
+        if save_to_h5:
+            xcc._init_h5(h5_f, iso_regions_f)
```
MRossol, line 122: Move to class __init__

reVX/least_cost_xmission/cost_creator.py
```diff
@@ -0,0 +1,304 @@
...
+
+            costs_arr = blc_arr * mults_arr
+            if save_to_h5:
+                xcc._write_layer(h5_f, f'tie_line_costs_{capacity}MW',
+                                 xcc._profile, costs_arr)
+            else:
+                xcc._write_geotiff(f'out/tif_line_costs{capacity}MW.tif',
+                                   xcc._profile, costs_arr)
+
+    def create_multipliers(self):
```
MRossol, line 143: silly, but how about `compute_multipliers` instead of create?
Also I think we should add two kwargs here:
`save_layer`
`save_geotiff`
and add the functionality to do the above into this method, then in the class method we would turn save_layer on and pass through the save_geotiff kwarg

reVX/least_cost_xmission/cost_creator.py
```diff
@@ -0,0 +1,304 @@
...
+            slope_mult = self._create_slope_mult(r_slope, d_mults['slope'])
+            mults_arr[default_mask] = mults_arr[default_mask] * slope_mult
+
+        # Set water multiplier last so we don't get super high multipliers at
+        # water body boundaries next to steep slopes
+        mults_arr[self._land_use == WATER_NLCD_CODE] = WATER_MULT
+
+        return mults_arr
+
+    def create_base_line_costs(self, capacity):
```
MRossol, line 198: Same as above..
silly, but how about `compute_*` instead of create?
Also I think we should add two kwargs here:
`save_layer`
`save_geotiff`
and add the functionality to do the above into this method, then in the class method we would turn save_layer on and pass through the save_geotiff kwarg

MRossol (changes requested):
This is looking great. Some general pythonic comments and suggestions.

Two other asks:
1) Lets get some test data started, let me know if I can help getting the landuse layer added to the test 'ri_exclusions.h5' file
2) Can we remove the old modules/methods that this will replace?

reVX/least_cost_xmission/config/xmission_config.py
```diff
@@ -0,0 +1,63 @@
...
+            new_substation_costs_f = os.path.join(D_DIR,
+                                                  'new_substation_costs.json')
+        with open(new_substation_costs_f, 'rt') as f:
+            self.new_sub_costs = json.load(f)
+
+        if upgrade_sub_costs_f is None:
+            upgrade_sub_costs_f = os.path.join(D_DIR,
+                                               'upgrade_substation_costs.json')
+        with open(upgrade_sub_costs_f, 'rt') as f:
+            self.upgrade_sub_costs = json.load(f)
```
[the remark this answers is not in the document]
mikebannis: Done

reVX/least_cost_xmission/config/xmission_config.py
```diff
@@ -0,0 +1,63 @@
+import os
+import json
+
+D_DIR = os.path.dirname(os.path.realpath(__file__))
+
+
+class XmissionConfig:
+    """ Load default tie lines configuration """
+
+    def __init__(self, power_to_voltage_f=None, base_line_costs_f=None,
```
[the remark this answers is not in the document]
mikebannis: Done

reVX/least_cost_xmission/config/xmission_config.py
```diff
@@ -0,0 +1,63 @@
...
+                 iso_mults_f=None, transformer_costs_f=None,
+                 iso_lookup_f=None, power_classes_f=None,
+                 min_power_classes_f=None, new_substation_costs_f=None,
+                 upgrade_sub_costs_f=None):
+        """
+        TODO
+        """
+        if power_to_voltage_f is None:
+            power_to_voltage_f = os.path.join(D_DIR, 'power_to_voltage.json')
+        with open(power_to_voltage_f, 'rt') as f:
```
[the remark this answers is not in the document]
mikebannis: done

reVX/least_cost_xmission/cost_creator.py
```diff
@@ -0,0 +1,304 @@
...
+
+        self._profile['dtype'] = 'float32'
+        self._profile['nodata'] = None
+
+        self._land_use_classes = land_use_classes
+        self._iso_mults = iso_mults
+        self._base_line_costs = base_line_costs
+        self._iso_lookup = iso_lookup
+        self._default_mults = default_mults
+        self._cell_size = CELL_SIZE
```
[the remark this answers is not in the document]
mikebannis: Added

reVX/least_cost_xmission/cost_creator.py
```diff
@@ -0,0 +1,304 @@
...
+    """
+    Class to create and save Transmission cost layers to a .h5 Exclusion file
+    - dist_to_coast (for valid sc_points)
+    - base_costs
+    - multiplier_*mw
+    - xmission_barrier
+    """
+    def __init__(self, iso_regions_f, land_use_classes, iso_mults,
+                 base_line_costs, iso_lookup, default_mults,
+                 excl_h5='/shared-projects/rev/exclusions/ATB_Exclusions.h5',
```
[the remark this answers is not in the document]
mikebannis: done

reVX/least_cost_xmission/cost_creator.py
```diff
@@ -0,0 +1,304 @@
...
+                            power_classes_f=power_classes_f)
+
+        xcc = cls(iso_regions_f, land_use_classes, xc.iso_mults,
+                  xc.base_line_costs, xc.iso_lookup, default_mults)
+        mults_arr = xcc.create_multipliers()
+
+        if save_to_h5:
+            xcc._init_h5(h5_f, iso_regions_f)
+        else:
+            xcc._write_geotiff('out/multipliers.tif', xcc._profile, mults_arr)
```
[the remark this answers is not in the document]
mikebannis: done

reVX/least_cost_xmission/cost_creator.py
mikebannis: done

reVX/least_cost_xmission/cost_creator.py
mikebannis: done

reVX/least_cost_xmission/cost_creator.py
mikebannis: done, though the kwargs get used before init, so a little out of order

reVX/least_cost_xmission/cost_creator.py
mikebannis: done

reVX/least_cost_xmission/cost_creator.py
mikebannis: done, need to use self._profile instead of iso_regions_f to init

reVX/least_cost_xmission/cost_creator.py
```diff
@@ -0,0 +1,304 @@
...
+        base_line_costs_f : None | str
+            Base line cost per mile for ISOs file. Use default if None
+        iso_lookup_f : None | str
+            JSON file linking ISO region code to name. Use default if None
+        power_classes_f : None | str
+            JSON of line capacities to use for reV power classes. Use default
+            if None
+        default_mults : dict
+            Multipliers for regions not specified in iso_mults_f.
+        save_to_h5 : bool
```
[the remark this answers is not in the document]
mikebannis: Has option for either/both/neither

reVX/least_cost_xmission/config/xmission_config.py
```diff
@@ -0,0 +1,63 @@
...
+class XmissionConfig:
+    """ Load default tie lines configuration """
+
+    def __init__(self, power_to_voltage_f=None, base_line_costs_f=None,
+                 iso_mults_f=None, transformer_costs_f=None,
+                 iso_lookup_f=None, power_classes_f=None,
+                 min_power_classes_f=None, new_substation_costs_f=None,
+                 upgrade_sub_costs_f=None):
+        """
+        TODO
```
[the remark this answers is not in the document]
mikebannis: There's a fine line between waiting for the args to stabilize before writing the doc string and writing the doc string before you forget what they are

reVX/least_cost_xmission/cost_creator.py
mikebannis: done

reVX/least_cost_xmission/cost_creator.py
```diff
@@ -46,113 +45,132 @@ def __init__(self, iso_regions_f, land_use_classes, iso_mults,
...
-    def create_multipliers(self):
+
+            if save_layer:
+                xcc.save_layer(f'tie_line_costs_{capacity}MW', costs_arr)
+
+            if save_geotiff:
+                xcc.create_geotiff(f'tie_line_costs{capacity}MW.tif',
+                                   costs_arr)
+
+    def compute_multipliers(self, slope_lyr, nlcd_lyr, land_use_classes,
```
MRossol, line 144: have the excl_h5 file be an arg here

MRossol (changes requested):

reVX/least_cost_xmission/cost_creator.py
MRossol: See new comment, I think this is right

reVX/least_cost_xmission/cost_creator.py
MRossol: close enough

reVX/least_cost_xmission/config/xmission_config.py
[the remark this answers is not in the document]
MRossol: I always write the docstrings when I create the function, at least I get the initial args/kwargs. That said I'm often guilty of forgetting to update them during changes/refactors. Still think its easier to do it when its fresh on the brain...

reVX/least_cost_xmission/cost_creator.py
```diff
@@ -46,113 +45,132 @@ def __init__(self, iso_regions_f, land_use_classes, iso_mults,
...
-        self._base_line_costs = base_line_costs
         self._iso_lookup = iso_lookup
-        self._default_mults = default_mults
-        self._cell_size = CELL_SIZE
+
+        if save_layer:
+            # TODO - use self._profile instead of regions file
+            self._init_h5(h5_fpath, iso_regions_fpath)
+
+        self._tiff_dir = tiff_dir
```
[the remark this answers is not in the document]
mikebannis: Is the problem saving the tiff path as an instance attribute? Should `create_geotiff()` not be in `build_cost_rasters()` at all? I need tiffs for development purposes, but I can rip them out once this code is stable.

reVX/least_cost_xmission/cost_creator.py
```diff
@@ -46,113 +45,132 @@ def __init__(self, iso_regions_f, land_use_classes, iso_mults,
...
         self._profile['nodata'] = None
 
-        self._land_use_classes = land_use_classes
-        self._iso_mults = iso_mults
-        self._base_line_costs = base_line_costs
         self._iso_lookup = iso_lookup
-        self._default_mults = default_mults
-        self._cell_size = CELL_SIZE
+
+        if save_layer:
```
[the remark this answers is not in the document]
mikebannis: I'll make saving to h5 mandatory once development of this file is complete or once the code checks if the h5 exists.

reVX/least_cost_xmission/cost_creator.py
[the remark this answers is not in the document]
MRossol: no, no, its that tiff_dir is ONLY used in `create_geotiff`, so it shouldn't be a class attribute. Instead it should be a kwarg in `build_cost_rasters` and then the actual geotiff path should be passed into `create_geotiff`.

What i would actually do is convert the `save_geotiff` kwargs into `tiff_dir` and use this logic:
```
if tiff_dir is not None:
    save_geotiff = True
else:
    save_geotiff = False
```

reVX/least_cost_xmission/cost_creator.py
```diff
@@ -46,113 +45,132 @@ def __init__(self, iso_regions_f, land_use_classes, iso_mults,
...
+        super().__init__(h5_fpath)
 
-        with ExclusionLayers(excl_h5) as el:
-            self._slope = el[slope_lyr]
-            self._land_use = el[nlcd_lyr]
-            # self._write_geotiff('out/slope.tif', self._profile, self._slope)
-            # self._write_geotiff('out/land_use.tif', self._profile, self._land_use)
+        logger.debug('Loading ISO regions')
+        with Geotiff(iso_regions_fpath) as gt:
+            self._iso_regions = gt.values[0]
```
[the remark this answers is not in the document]
mikebannis: Evan R made it sound like properly creating the reV projection is hard (I screwed it up already, hence that small shift we saw earlier). Pulling the profile from the atb exclusions file seemed more reliable than pulling it from a rasterized shapefile.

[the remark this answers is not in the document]
MRossol: I think this is a great example of easier != right. If you look at ExclusionsConverter you'll see there is a lot of logic when adding layers to make sure they match the projection. If our iso geotiff projection is different than the other layers we need to fix that first.

reVX/least_cost_xmission/cost_creator.py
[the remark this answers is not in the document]
mikebannis: What about 
```
if tiff_dir:
    xcc.create_geotiff('mults.tif', tiff_dir, mults_arr)
```
Less code, though using to `tiff_dir` to  set `save_geotiff` is more explicit.

reVX/least_cost_xmission/cost_creator.py
mikebannis: Done

reVX/least_cost_xmission/cost_creator.py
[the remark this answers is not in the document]
mikebannis: Done

reVX/least_cost_xmission/cost_creator.py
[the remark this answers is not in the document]
mikebannis: Sounds good to me, and I'll get that fixed. FYI, here's the warning from the ExclusionsConverter with the bad raster:

DEBUG - 2021-07-23 13:23:43,649 [exclusions_converter.py:307] : 	- Unique profile for tie_line_costs_102MW added:
{'driver': 'GTiff', 'dtype': 'float32', 'nodata': None, 'width': 48640, 'height': 33792, 'count': 1, 'crs': '+proj=aea +lat_0=40 +lon_0=-96 +lat_1=20 +lat_2=60 +x_0=0 +y_0=0 +ellps=GRS80 +units=m +no_defs=True', 'transform': (90.0, 0.0, -2245497.13, 0.0, -90.0, 1338250.676), 'blockxsize': 128, 'blockysize': 128, 'tiled': True, 'compress': 'lzw', 'interleave': 'band'}

reVX/least_cost_xmission/cost_creator.py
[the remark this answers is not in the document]
MRossol: the if statement logic is fine, but I don't follow the `create_geotiff` logic. Passing in the name and directory is silly, just pass in the full file path and create it from `tiff_dir` and `mults.tif` in the classmethod. Also when ever generating filenames be verbose: `tie-line_cost_multipliers.tiff`

reVX/least_cost_xmission/cost_creator.py
```diff
@@ -0,0 +1,410 @@
...
+        layer_name : str
+            Name of layer to save in h5
+        data : np.ndarray
+            Data to savein h5
+        """
+        logger.debug(f'Saving {layer_name} to h5')
+        self._profile['nodata'] = None
+        self._write_layer(self._excl_h5, layer_name, self._profile, data)
+
+    def create_geotiff(self, tiff_dir, geotiff_fpath, data):
```
MRossol, line 305: Simplify this to be just:
```
def create_geotiff(self, geotiff_fpath, data):
```
There is no need to do an os.path.join inside this method.

reVX/least_cost_xmission/cost_creator.py
```diff
@@ -0,0 +1,410 @@
...
+
+        for power_class, capacity in xc['power_classes'].items():
+            logger.info(f'Calculating costs for class {power_class} using a '
+                        f'{capacity}MW line')
+            blc_arr = xcc.compute_base_line_costs(capacity,
+                                                  xc['base_line_costs'],
+                                                  cell_size)
+
+            if save_geotiff:
+                xcc.create_geotiff(tiff_dir, f'blc_{capacity}MW.tif', blc_arr)
```
MRossol, line 135: change `blc` to `base_line_cost`

reVX/least_cost_xmission/cost_creator.py
```diff
@@ -0,0 +1,410 @@
...
+
+        if land_use_classes is None:
+            land_use_classes = NLCD_LAND_USE_CLASSES
+
+        mults_arr = xcc.compute_multipliers(input_h5_fpath, slope_lyr,
+                                            nlcd_lyr, land_use_classes,
+                                            xc['iso_mults'], default_mults)
+
+        if save_geotiff:
+            xcc.create_geotiff(tiff_dir, 'multipliers.tif', mults_arr)
```
MRossol, line 125: how about `tie_line_multipliers.tif`

reVX/least_cost_xmission/cost_creator.py
```diff
@@ -0,0 +1,410 @@
...
+            assert isinstance(values, list)
+
+            for value in values:
+                index = land_use == value
+                indices.append((index, multiplier, value))
+
+        for i in indices:
+            mult_raster[i[0]] = i[1]
+
+        return mult_raster
```
MRossol, line 410: Lets add a second class method, could be `run`, `build`, or `create` that would
1) run `build_test_costs`
2) add any extra layers using [`layers_to_h5`](https://github.com/NREL/reVX/blob/adab7d9d2e64ff06cf4ad8438f17952004f34e5d/reVX/utilities/exclusions_converter.py#L508)

tests/test_x_mission_costs.py
```diff
@@ -6,30 +6,37 @@
...
+    assert round(PathFinder._calc_path_cost(costs, i2), 5) == 17.27817
 
 
 def test_land_use_multiplier():
     """ Test land use multiplier creation """
     lu_mults = {'forest': 1.63, 'wetland': 1.5}
     arr = np.array([[[0, 95, 90], [42, 41, 15]]])
-    cm = CostMultiplier()
-    out = cm._create_land_use_mult(arr, lu_mults)
+    xcc = XmissionCostCreator(INPUT_H5F, ISO_REGIONS_F, {})
```
MRossol, line 38: whats up with the empty dict here?

MRossol (changes requested):

MRossol:
@mikebannis  getting super close! This is looking great. A few minor comments

MRossol:
one more comment, lets split the test file into two:
1) test_xmission_cost_creator.py
2) test_xmission_least_cost.py

MRossol:
@mikebannis wanted to check in on this PR, would be nice to wrap this up before the next one...

mikebannis:
Working on the path finding code. I'll wrap this up tonight or tmw

MRossol:
No rush just a gentle nudge before we both forget what we were thinking, so definitely tomorrow, you don't get paid enough to work tonight...

tests/test_x_mission_costs.py
mikebannis: iso region dict. It's not needed for the test.

reVX/least_cost_xmission/cost_creator.py
mikebannis: done

reVX/least_cost_xmission/cost_creator.py
mikebannis: done

reVX/least_cost_xmission/cost_creator.py
mikebannis: Keeping os.path.join inside create_geotiff is much DRYer and keeps the code in build_cost_rasters() shorter. I'd prefer to save the tiff_dir as an attribute to further simply the function arguments. Thoughts?

reVX/least_cost_xmission/config/xmission_config.py
```diff
@@ -0,0 +1,88 @@
...
+        min_power_classes_fpath : None | str
+            JSON file with actual line capacities to use for reV power classes.
+        new_sub_costs_fpath : None | str
+            JSON file with existing new substation construction costs
+        upgrade_sub_costs_fpath : None | str
+            JSON file with existing substation upgrade costs
+        """
+        super().__init__()
+
+        if power_to_voltage_fpath is None:
```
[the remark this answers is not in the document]
mikebannis: Noted. Let's revisit this once the code works.

reVX/least_cost_xmission/cost_creator.py
```diff
@@ -23,21 +25,18 @@ class XmissionCostCreator(ExclusionsConverter):
     - multiplier_*mw
     - xmission_barrier
     """
-    def __init__(self, iso_regions_f, land_use_classes, iso_mults,
-                 base_line_costs, iso_lookup, default_mults,
-                 excl_h5='/shared-projects/rev/exclusions/ATB_Exclusions.h5',
-                 slope_lyr='srtm_slope', nlcd_lyr='usa_mrlc_nlcd2011'):
+    def __init__(self, h5_fpath, iso_regions_fpath, atb_excl_h5, iso_lookup,
```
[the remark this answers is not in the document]
mikebannis: Renamed.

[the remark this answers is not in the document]
mikebannis: I'm ignoring the profile issue for a bit.

reVX/least_cost_xmission/cost_creator.py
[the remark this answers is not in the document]
mikebannis: See notes below.

reVX/least_cost_xmission/cost_creator.py
mikebannis: What do you mean by extra layers?

reVX/least_cost_xmission/cost_creator.py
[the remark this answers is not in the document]
MRossol: LOL, still waiting on Evan to fix that?

reVX/least_cost_xmission/cost_creator.py
MRossol: Transmission barrier layer and dist_to_coast layers I think are the only "other" layers we need. i.e none cost layers

tests/test_x_mission_costs.py
MRossol: Ah! Should we make that a kwarg then?

reVX/least_cost_xmission/cost_creator.py
MRossol: I don't fully follow. Since this is a public method it seem much more straight forward to just request the full geotiff path and data to write (for anyone using this class that is not us). Then in the `build_cost_rasters` class method the tiff_dir is an input value so its a local variable, and we can create the custom file paths with `os.path.join` inside the `if save_geotfif` logic on lines 135 and 142.

reVX/least_cost_xmission/cost_creator.py
```diff
@@ -0,0 +1,410 @@
...
+            costs_arr = blc_arr * mults_arr
+
+            xcc.save_layer(f'tie_line_costs_{capacity}MW', costs_arr)
+
+            if save_geotiff:
+                xcc.create_geotiff(tiff_dir, f'tie_line_costs{capacity}MW.tif',
+                                   costs_arr)
+
+    @classmethod
+    def build_test_costs(cls, h5_fpath, iso_regions_fpath, input_h5_fpath,
```
MRossol, line 146: This isn't useful outside of the tests, we should just move this code to the tests as a function that can be run if needed.

reVX/least_cost_xmission/cost_creator.py
```diff
@@ -0,0 +1,410 @@
...
+        """
+        nodata = self._profile['nodata']
+        if tiff_dir != '':
+            geotiff_fpath = os.path.join(tiff_dir, geotiff_fpath)
+        logger.debug(f'Saving {geotiff_fpath}')
+        self._write_geotiff(geotiff_fpath, self._profile, data)
+
+        # _write_geotiff sets nodata to the max for the dtype. For float32 this
+        # then breaks JSONifying the profile when writing h5s. Reset it.
+        self._profile['nodata'] = nodata
```
MRossol, line 326: Why is nodata pulled out and then added back to the profiel? I dont' see it being used in this function, are the above comments supposed to be a TODO?

MRossol (changes requested):

reVX/least_cost_xmission/config/__init__.py
```diff
@@ -4,7 +4,7 @@
 """
 import os
 
-from .xmission_config import XmissionConfig
+from reVX.least_cost_xmission.config.xmission_config import XmissionConfig
```
mikebannis, line 7: Out of curiosity, why the absolute versus relative import? It seems the relative import is less work if the folder structure changes.

reVX/least_cost_xmission/config/config.json
```diff
@@ -0,0 +1 @@
...C", "land_use": {"cropland": 1, "forest": 2.25, "suburban": 1.3, "urban": 1.59, "wetland": 1.2}, "slope": {"hill_mult": 1.4, "hill_slope": 2, "mtn_mult": 1.75, "mtn_slope": 8}}, {"iso": "SCE", "land_use": {"cropland": 1, "forest": 3, "suburban": 2, "urban": 3, "wetland": 2}, "slope": {"hill_mult": 1.5, "hill_slope": 2, "mtn_mult": 2, "mtn_slope": 8}}, {"iso": "MISO", "land_use": {"cropland": 1.03, "forest": 1.2, "suburban": 1.08, "urban": 1.2, "wetland": 1.8}, "slope": {"hill_mult": 1.1, "hill_slope": 2, "mtn_mult": 1.21, "mtn_slope": 8}}, {"iso": "Southeast", "land_use": {"cropland": 1, "forest": 1.47, "suburban": 1.75, "urban": 1.12, "wetland": 1.33}, "slope": {"hill_mult": 1.2, "hill_slope": 2, "mtn_mult": 1.6, "mtn_slope": 8}}], "base_line_costs": {"MISO": {"102": 1255000, "1500": 2787000, "205": 1446000, "400": 1695000}, "SCE": {"102": 1524000, "1500": 4777000, "205": 2084000, "400": 3034000}, "Southeast": {"102": 819000, "1500": 4056000, "205": 984000, "400": 1568000}, "TEPPC": {"102": 984000, "1500": 2248000, "205": 1180000, "400": 1570000}}, "nlcd_land_use_classes": {"cropland": [80, 81], "forest": [41, 42, 43], "wetland": [90, 95], "suburban": [21, 22, 23], "urban": [24]}}
```
mikebannis, line 1: Is it safe to assume github actions will format that?

MRossol: Yeah I did a no-verify commit just to get up before calling it a day. Going to finish it up / clean it up now.

reVX/least_cost_xmission/config/__init__.py
MRossol: No good reason beyond Grant's preference that we started following in reV and has bleed over to reVX. We have run into some strange import issues in the past but I think that was mainly noobness + python 3.6 being weird.

mikebannis:
I merged xmission/cost_layers into xmission/cost_paths on my computer. Should we just close this and delete the PR?
At this point I'm tempted to just delete mjb/feat/x...

MRossol:
Fine by me, i'll close this PR with out merging into mjb/.... and agreed lets delete mjb/... I'll delete it here so you just have to delete your remote branch

PR #125, opened 2021-08-13
add new getters to config

MRossol:
Add a new getitem to config to pull the inverse dicts
add cap to v converters and substation and xformer cost getters methods


reVX/least_cost_xmission/cost_creator.py
```diff
@@ -266,7 +272,7 @@ def compute_base_line_costs(self, capacity, base_line_costs):
             logger.info(f'Processing costs for {iso} for {capacity}MW')
             iso_code = self._iso_lookup[iso]
             cost_per_mile = base_line_costs[iso][str(capacity)]
-            cost_per_cell = cost_per_mile / METERS_IN_MILE * self.PIXEL_SIZE
+            cost_per_cell = cost_per_mile / 1609.344 * self.PIXEL_SIZE
```
mikebannis, line 275: I do like METERS_IN_MILE because i don't know what 1609.344 is

reVX/least_cost_xmission/cost_creator.py
```diff
@@ -239,7 +245,7 @@ def compute_multipliers(self, iso_mults, excl_h5=None,
 
         # Set water multiplier last so we don't get super high multipliers at
         # water body boundaries next to steep slopes
-        mults_arr[land_use == WATER_NLCD_CODE] = WATER_MULT
+        mults_arr[land_use == 11] = 10
```
mikebannis, line 248: The 10 should probably be a kwarg. Using WATER_NLCD_CODE is a bit more obvious than 11, though the meaning can be derived pretty quickly from the comment. I'm indifferent.

mikebannis:
Looks good

reVX/least_cost_xmission/cost_creator.py
mikebannis: It would actually be nice to change PIXEL_SIZE to PIXEL_SIZE_METERS

PR #130, opened 2021-08-31
Base LeastCostPath class for BA-BA paths

MRossol:


mikebannis (approved):

reVX/least_cost_xmission/least_cost_paths.py
```diff
@@ -0,0 +1,383 @@
...
+        features_fpath : str
+            Path to geopackage with transmission features
+        resolution : int, optional
+            SC point resolution, by default 128
+        xmission_config : str | dict | XmissionConfig, optional
+            Path to Xmission config .json, dictionary of Xmission config
+            .jsons, or preloaded XmissionConfig objects, by default None
+        """
+        self._check_layers(cost_fpath)
+        self._config = TieLineCosts._parse_config(
```
mikebannis, line 48: Is calling the private function intentional?

MRossol: Yeah, i probably over coded this. the _parse_config method in TieLineCosts checks for a path or a XmissionConfig object, and if needed creates the XmissionConfig object. We could just require TieLineCosts to take a XmissionConfig object and move the method up to this class, but also it runs instantaneously...

PR #142, opened 2022-03-31
Parcel setbacks

ppinchuk:
Implemented Parcel setbacks using negative buffers. Allowed modeling discrete, existing ordinances on a per county basis (w/an input CSV)  as well as a generic setback across the nation using static or height based setbacks.

The result, before rasterization, looks something like:

image

The blue regions are the input parcels while the red regions are the exclusion areas.

Also renamed `wind-setbacks` to `setbacks`, including all files and the cli. Since the cli was updated, I increased the minor version number as well.


ppinchuk:
I think the rasterized output looks good as well. Here is Illinois  using the setback CSV that Anthony provided:

image

And here is that inset region with more zoom:

image

reVX/setbacks/base.py
```diff
@@ -0,0 +1,658 @@
...
+        Parameters
+        ----------
+        excl_fpath : str
+            Path to .h5 file containing exclusion layers, will also be the
+            location of any new setback layers
+        plant_height : float | int
+            Plant height (m), used to determine setback distance.
+        regulations_fpath : str | None, optional
+            Path to regulations .csv file, if None create generic
+            setbacks using max-tip height * "multiplier", by default None
```
grantbuster, line 41: Can we describe what this regulations_fpath csv should look like? What are the required columns and what does each row represent? County, State, Value Type, Value? Does it need the FIPS fields or is that optional? Even an incomplete description would be useful.

reVX/setbacks/solar_setbacks.py
```diff
@@ -0,0 +1,219 @@
...
+            (geom, 1) for geom in setback_features.buffer(0).difference(
+                setback_features.buffer(-1 * self.generic_setback)
+            )
+        ]
+
+        return self._rasterize_setbacks(setbacks)
+
+    @staticmethod
+    def _compute_local_setbacks(features, cnty, setback):
+        """Compute local features setbacks.
```
grantbuster, line 48: For all the "local" and "generic" setback methods, can we clarify that local is with a county-specific regulations file and generic is based on the base height * multiplier?

reVX/setbacks/solar_setbacks.py
```diff
@@ -0,0 +1,219 @@
...
+        return regulations
+
+    def _check_regulations(self, features_fpath):
+        """
+        Reduce regs to state corresponding to features_fpath if needed.
+
+        Parameters
+        ----------
+        features_fpath : str
+            Path to shape file with features to compute setbacks from.
```
grantbuster, line 106: Needs to have the state in the filename?

reVX/setbacks/base.py
```diff
@@ -0,0 +1,658 @@
...
+        -------
+        setback : float | None
+            setback distance in meters, None if the setback "Value Type"
+            was not recognized
+        """
+
+        setback_type = county_regulations['Value Type'].strip()
+        setback = float(county_regulations['Value'])
+        if setback_type == 'Maximum Structure Height':
+            setback *= self.plant_height
```
grantbuster, line 381: if the setback type is maximum structure height, isnt the setback just equivelant to the plant height? I would think we only multiply by the plant height if value type is "Max-tip Height Multiplier" not "Maximum Structure Height"

reVX/setbacks/base.py
```diff
@@ -0,0 +1,658 @@
...
+    def __init__(self, excl_fpath, plant_height, regulations_fpath=None,
+                 multiplier=None, hsds=False, chunks=(128, 128)):
+        """
+        Parameters
+        ----------
+        excl_fpath : str
+            Path to .h5 file containing exclusion layers, will also be the
+            location of any new setback layers
+        plant_height : float | int
+            Plant height (m), used to determine setback distance.
```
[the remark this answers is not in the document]
ppinchuk: I guess I just mimicked Anthony's nomenclature: "C) include ability to model static or height based setbacks (e.g., 30-meters or 1.1x height of plant).". I even asked him recently what plant height I should use for a test run and he didn't mention anything about me using the wrong term (maybe he was just being polite lolol). 

I think plant height also matches the user config nicely. The wind one looks something like
```
...
"hub_height": 135,
"rotor_diameter": 200
...
```

while the solar one would look like 

```
...
"plant_height": 5
...
```

But in any case, I'm not tied to one way or another. I will say that changing the name to `base_setback` would definitely require some effort, as the cli and config file will also have to be updated. That being said, its totally doable, so let me know if you think that's worth it and I will work on that.

reVX/setbacks/base.py
ppinchuk: So I explicity checked with Anthony and Yuying, and they said that for the solar csv file, a value type of `Maximum Structure Height` represents a multipler. I can totally see how that would be confusing in the base class however, so I will move that to the solar setbacks subclass and keep this one generic.

reVX/setbacks/solar_setbacks.py
ppinchuk: Yup, done!

reVX/setbacks/solar_setbacks.py
ppinchuk: Yup, added that to docstring!

reVX/setbacks/base.py
ppinchuk: Oops still WIP

reVX/setbacks/base.py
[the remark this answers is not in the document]
grantbuster: I think it might be worth it... In this case the height of the solar panel is not 5 meters, you're just modeling a 5 meter setback for the plant. Same with the 30m/1.1x example. The height of the plant is not 30 meters, its just a generic 30 meter setback.

reVX/setbacks/base.py
grantbuster: Haha yeah, definitely not your fault but reading through i've been asking myself what the hell are these variables?? We can't totally prevent the inheritance of technical debt but we can try and improve it. Thanks!

reVX/setbacks/base.py
grantbuster: Yeah might be a good place for an abstract method. That is super confusing about the multiplier.... we should touch base with anthony about nomenclature.

reVX/setbacks/base.py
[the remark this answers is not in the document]
grantbuster: I may be wrong about this, let's confirm with anthony

ppinchuk:
One of these days I will beat the linter... just you wait

reVX/setbacks/base.py
ppinchuk: Good point, I will throw that in

grantbuster:
haha it gets easier i promise!

reVX/setbacks/base.py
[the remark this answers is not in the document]
ppinchuk: So where do we stand on this?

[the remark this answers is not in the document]
grantbuster: okay can you set this variable to `base_setback_dist`? and also change the entry in the solar config? Painful but necessary IMO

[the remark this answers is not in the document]
ppinchuk: On it!

reVX/setbacks/base.py
grantbuster: Okay here's the plan: lets do the following options (plus any that you think are necessary) for solar: "meters" and "Structure Height Multipler". We should also do the following:

1. Make `_get_setback()` -> `get_regulation_setback()`
2. Add in the docstrings what the options are (Max-tip height multiplier, meters, etc...) No need for description, these should be self-documenting 
3. Make this `setback_type.lower()` so its not case-sensitive

ppinchuk: OK, I think I got everything but definitely let me know if I've overlooked anything

grantbuster (approved):

PR #145, opened 2022-06-03
Parcel setback refactor

ppinchuk:
Minor refactor in preparation for a larger update for water setbacks


ppinchuk:
Messed up something here with the branches... working on a fix

ppinchuk:
Ok should be good now

PR #150, opened 2022-06-07
Feat/offshore lcp

mikebannis:
Not done yet and need to do a lot of linting. Just leaving this here as a heads up and to see if there are any grievous issues I need to fix.


reVX/config/least_cost_xmission.py
```diff
@@ -109,6 +110,50 @@ def __init__(self, config):
...
+        -------
+        _name : str
+            Job name.
+        """
+        if self._name is None:
+            if self.get('name') is not None:
+                self._name = self.get('name')
+            else:
+                # name defaults to base directory name
+                self._name = os.path.basename(os.path.normpath(self.dirout))
```
grantbuster, line 129: you should do something like this:
```
default_name = os.path.basename(os.path.normpath(self.dirout))
self._name = self.get('name', default_name)

File: reVX/least_cost_xmission/aoswt_utilities.py [hunk 2]
```diff
@@ -0,0 +1,604 @@
+"""
+Various utility functions to prep data for AOSWT processing.
```
grantbuster, line 2: Would this module not work for the pacific? Why not?

reVX/least_cost_xmission/aoswt_utilities.py
```diff
@@ -0,0 +1,604 @@
...
+    for pt in pts.geometry:
+        end = Point(pt.x + 50, pt.y + 50)
+        line = LineString([pt, end])
+        new_geom.append(line)
+    lines = pts.set_geometry(new_geom, crs=crs)
+
+    # Append some fake values to make the LCP code happy
+    lines['ac_cap'] = 9999999
+    lines['category'] = 'Substation'
+    lines['voltage'] = 69
```
grantbuster, line 61: maybe make the fake voltage 0 or 1e9?

reVX/least_cost_xmission/aoswt_utilities.py
```diff
@@ -0,0 +1,604 @@
...
+    pois.to_file(out_f, driver="GPKG")
+    print('Complete')
+
+
+class CombineRasters:
+    """
+    Combine layers to create composite friction and barrier rasters. Merge
+    with existing land cost and barriers and save to h5.
+    """
+    OS_FRICTION_FNAME = 'os_friction.tif'
```
grantbuster, line 92: What is OS? Offshore? Onshore is also OS?

reVX/least_cost_xmission/aoswt_utilities.py
```diff
@@ -0,0 +1,604 @@
...
+            where ras_valX is the value in the raster, and fric_valX is the
+            desired friction values. Any values in the raster that are not
+            specified in the dict are assumed to have no friction. 'fname.tif'
+            is the file name of the raster.
+        slope_file : str, optional
+            Path to slope friction tiff
+        save_tiff : bool, optional
+            Save composite friction to tiff if true
+        """
+        print('Loading friction layers')
```
grantbuster, line 187: Please convert all print statements to logger.info() statements

reVX/least_cost_xmission/aoswt_utilities.py
```diff
@@ -0,0 +1,604 @@
...
+                'width': ras.width,
+                # 'dtype': ras.dtype,
+                'count': 1,
+                'compress': 'lzw'
+            }
+        return profile
+
+    def _load_layer(self, f, val, verbose=False):
+        """
+        Load a layer from a tiff and set appropriate cells to 1
```
grantbuster, line 554: Can you only handle tiff inputs? Should we support our collection of h5 exclusion layers?

grantbuster (changes requested):
Most of this looks good but its hard to understand exactly what your goals are, i'd appreciate a quick call just to get the rundown of what all this work is about.

reVX/least_cost_xmission/aoswt_utilities.py
mikebannis: The substation needs a transmission line with adequate voltage for the code to connect to it. I arbitrarily chose 69kV as a standard voltage. 1e9 may work as well but I would need to check it.

mikebannis:
TODO:

- When saving paths, also export to CSV
- calculate % path distance offshore vs land and save to dist_offshore_frac column
- change README to rst
- remove AOSWT from naming when appropriate and make ocean agnostic
- Update linter config (done)
- Make 500kV default voltage for fake t-lines

ppinchuk:
@mikebannis Are there any plans to finalize this? I will be working on some changes to LCP next week, so I'd like to know if I need to work around this PR or not

ppinchuk:
After conversation with Mike, I will attempt to do bare minimum to make this PR mergeable

reVX/least_cost_xmission/aoswt_utilities.py
ppinchuk: Does not seem to be major demand for this request, so I propose we shelve it. In the immediate future, anyone looking to use this function with an h5 exclusion layer can use the [`$ exclusions layers-from-h5`](https://github.com/NREL/reVX/blob/3a3694a171ab672b35865a27ba2f64c470280dcc/reVX/cli.py#L212) command line utility to extract h5 layers as tiff.

reVX/least_cost_xmission/aoswt_utilities.py
ppinchuk: Done

reVX/least_cost_xmission/aoswt_utilities.py
ppinchuk: Renamed to "Offshore" for clarity

reVX/least_cost_xmission/aoswt_utilities.py
ppinchuk: Looks like Mike was in the process of working on this (see quote form him below), but the demand for this work dropped off. I propose to shelve this question until demand for this work arises once more


> TODO:
> 
> * When saving paths, also export to CSV
> * calculate % path distance offshore vs land and save to dist_offshore_frac column
> * change README to rst
> * remove AOSWT from naming when appropriate and make ocean agnostic
> * Update linter config (done)
> * Make 500kV default voltage for fake t-lines

mikebannis: The AOSWT code would really work for determining paths anywhere within the reV raster, including on land.

ppinchuk: Should we rename the file?

mikebannis: The code was written for AOSWT (Atlantic off-shore wind transmission) and has only been used for AOSWT so far. Naming it after it provides some history, assuming the user knows what AOSWT is. Keeping it named AOSWT until it's used for something else makes sense to me. That said, I won't be offended if you decide to change it.

reVX/least_cost_xmission/aoswt_utilities.py
mikebannis: A little context: I'm not very good at using h5s and I frequently load raster and vector data into QGIS for using basemaps and zooming. QGIS doesn't speak h5. Tiffs are also much smaller than h5 files and easier to copy to and from eagle. So that's why some of this code is setup for using tiffs. I believe all the input files have already been converted to h5 files. Switching over to a pure h5 workflow should not be an issue.

grantbuster (approved):

PR #152, opened 2022-06-21
Pp/water setbacks

ppinchuk:
Added support for water setbacks

### DIFF
--- a/reVX/setbacks/setbacks_cli.py
+++ b/reVX/setbacks/setbacks_cli.py
@@ -481,20 +481,10 @@ def get_node_cmd(name, config):
         args.append('-v')
 
     feature_type = config.feature_type
-    if feature_type == 'structure':
-        args.append('structure-setbacks')
-    elif feature_type == 'road':
-        args.append('road-setbacks')
-    elif feature_type == 'transmission':
-        args.append('transmission-setbacks')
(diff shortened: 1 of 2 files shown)


reVX/setbacks/water_setbacks.py
```diff
@@ -0,0 +1,35 @@
...
+
+        Returns
+        -------
+        regulations : pandas.DataFrame
+            Parcel regulations table
+        """
+        regulations = super()._parse_regulations(regulations_fpath)
+
+        feats = regulations['Feature Type'].apply(str.strip).apply(str.lower)
+        regulations = regulations.loc[feats == 'water']
```
grantbuster, line 33: Nice, inheritance is pretty powerful!

grantbuster (approved):

reVX/setbacks/water_setbacks.py
mikebannis: Minor and late to the party: `feats = regulations['Feature Type'].str.strip().str.lower()` is an alternative to `apply()`

ppinchuk: Oh dang, didn't know about that shorthand. Will keep it in mind moving forward!

Issue #146, opened 2022-06-06
Method to Merge Existing and Extrapolated Ordinances


alopez04:
**Why this feature is necessary:**
We have methods to either A) create spatial exclusions for existing (real world via a CSV) ordinances or B) spatial exclusions for hypothetical exclusions for the entire country. Fantastic tools, however, our typical use case is to merge these two into a single exclusion layer. The single layer uses the hypothetical as the base, but then overrides where we have actual ordinances. 

**A possible solution is:**
Either add a flag to handle all of this in one go e.g., model existing ordinances and apply these elsewhere. Or have a new method that assumes you've already created the necessary datasets and just want to merge them (less ideal). 

**I have considered the following alternatives:**
We typically do this "by hand" as needed, but the demand has grown significantly and would be very helpful. 

**Additional context**

**Charge code**
WEPG.10348.08.01.06

**Urgency / Timeframe**
Not urgent (doing by hand atm) / July 2022

PR #154, opened 2022-06-30
Merge methods

ppinchuk:
Added support for merging existing and extrapolated ordinances in one `reVX` invocation.
No new flag was needed - the user can now simply provide *both* a setback multiplier and a regulations file.

Sample comparisons shown below (with extreme cases to illustrate the functionality):


tests/test_setbacks.py
```diff
@@ -372,13 +402,139 @@ def test_partial_exclusions_upscale_factor_less_than_1(mult):
...
+
+        local_layer = local_setbacks.compute_setbacks(features_path,
+                                                      max_workers=1)
+
+        merged_setbacks = setbacks_class(EXCL_H5, *setback_distance,
+                                         regulations_fpath=regs_fpath,
+                                         weights_calculation_upscale_factor=sf,
+                                         multiplier=100)
+        merged_layer = merged_setbacks.compute_setbacks(features_path,
+                                                        max_workers=1)
```
grantbuster, line 449: I'm a little confused on where the actual merge request happens... Like you're not actually requesting a merge of two exclusion layers but instead just combining the setback distance input with an explicit regulations file? Otherwise looks good to me!

grantbuster (approved):

tests/test_setbacks.py
ppinchuk: Right, I think the title is a bit misleading (I essentially copied the issue request)

[Anthony had said](https://github.com/NREL/reVX/issues/146#issue-1262533179):  "Either add a flag to handle all of this in one go e.g., model existing ordinances and apply [generic ordinances] elsewhere. Or have a new method that assumes you've already created the necessary datasets and just want to merge them (less ideal)."

He said the second option is less ideal so I went with the first idea (except there was no need for a new flag value)

ppinchuk: Merging this so that Anthony can start using it, but let me know if you have any other questions/objections!

PR #155, opened 2022-07-05
Separation of wind and solar setbacks for parcels and water

ppinchuk:
Had a chat with Anthony and he explained that both parcel and water setbacks have local ordinances that are based on turbine heights (e.g. rotor diameter multiplier, max tip-height multiplier, etc). Since these are also meant for use in solar, I split both the `ParcelSetbacks` and the `WaterSetbacks` classes into solar and wind components. Users can now specify either `base_setback_distance` or both `hub_height` and `rotor_diameter` in the config, and the code will automatically apply the solar and the wind setbacks, respectively.

Note that if Anthony and co. ever need solar versions of the the wind-only setbacks (structures, railroads, transmissions, etc), then it may be worthwhile to refactor the setbacks code entirely. The code was originally written for wind setbacks only, and splitting it into two components over time has led to increasingly (and, importantly, unnecessarily) complex code.


reVX/config/setbacks.py
```diff
@@ -41,6 +41,17 @@ def _preflight(self):
             logger.error(e)
             raise ConfigError(e)
 
+        no_base_setback = self.base_setback_dist is None
+        invalid_turbine_specs = (self.rotor_diameter is None
+                                 or self.hub_height is None)
+        not_enough_info = no_base_setback and invalid_turbine_specs
```
grantbuster, line 47: haha good variable names

grantbuster (dismissed):

ppinchuk:
Merging this b/c changes after previous approval were super minimal

PR #158, opened 2022-08-05
Gb/plexos fout

grantbuster:


reVX/plexos/base.py
```diff
@@ -561,3 +586,140 @@ def _ammend_output_meta(self, row_idx, sc_gids, res_gids, gen_gids,
...
+    @staticmethod
+    def get_unique_plant_names(table, name_col, tech_tag=None):
+        """Get a list of ordered unique plant names
+
+        Returns
+        -------
+        list | None
+        """
+        names = None
+        if name_col is not None:
```
ppinchuk, line 652: You could consider doing something like 
```
if name_col is None:
    return
```

That would allow you to remove 1 indent from the rest of the code :)

reVX/plexos/base.py
```diff
@@ -561,3 +586,140 @@ def _ammend_output_meta(self, row_idx, sc_gids, res_gids, gen_gids,
...
+                       .format(name_col, sorted(table.columns.values)))
+                logger.error(msg)
+                raise KeyError(msg)
+
+            names = table[name_col].values.tolist()
+
+            if tech_tag is not None:
+                names = [name + f' {tech_tag}' for name in names]
+
+            seen = set()
```
ppinchuk, line 665: You could consider using `Counter` here:

```
from collections import Counter
counter = Counter()
for i, name in enumerate(names):
    counter.update([name])
    count = counter[name]
    if count > 1:
        names[i] = name + " {}".format(count)
```

reVX/plexos/base.py
```diff
@@ -561,3 +586,140 @@ def _ammend_output_meta(self, row_idx, sc_gids, res_gids, gen_gids,
...
+            msg = ('Cannot use profiles that are not at least hourly! '
+                   'Received shape {}'.format(profiles.shape))
+            logger.error(msg)
+            raise RuntimeError(msg)
+
+        steps_per_hour = len(profiles) // 8760
+
+        # use jan 1 to avoid daylight savings
+        date = datetime.datetime(2011, 1, 1)
+        date = pytz.timezone(timezone).localize(date)
```
ppinchuk, line 622: Whoa this is cool, I have never heard of `pytz`. Looks super useful for these annoying timezone conversions. And actually, now that I think of it, maybe I can use it to replace that ugly utc offset dict in the `Cities` class in `ntps_load`. Will explore this!

ppinchuk (approved):
LGTM, just minor suggestions that can be ignored

reVX/plexos/base.py
grantbuster: Ah yeah i like this pattern, done

reVX/plexos/base.py
grantbuster: Cool i didnt know about counter! I implemented. Same number of lines i think but more robust and easy to read.

reVX/plexos/base.py
grantbuster: Yeah it's still painful but helps a little bit haha

PR #168, opened 2022-12-20
added option for explicit rpm cluster CRS with default to WGS 84

grantbuster:


reVX/rpm/rpm_clusters.py
```diff
@@ -322,13 +323,17 @@ def _get_cluster_geom(gdf_points):
...
+                           source_crs="EPSG:4326", target_crs=None):
         """
         Generate cluster polygons and save to shapefile
         """
         geometry = [Point(xy) for xy in zip(meta.longitude, meta.latitude)]
         gdf_points = gpd.GeoDataFrame(meta, geometry=geometry,
-                                      crs="EPSG:5070")
+                                      crs=source_crs)
+
+        if target_crs is not None:
```
ppinchuk, line 335: Do you know if Geopandas efficiently handles the case where the df crs == target crs? If we want to be extra cautious, we could change this to `if target_crs != source_crs:` in order to be completely sure that no unnecessary computation is happening

ppinchuk (approved):
LGTM

netr0m (dismissed):
LGTM!

reVX/rpm/rpm_clusters.py
grantbuster: No idea haha that's a good idea to add that!

ppinchuk (approved):

PR #169, opened 2022-12-22
Setbacks Multi-Geometry bug fix

ppinchuk:
If a county is broken up into multiple geometries (e.g. along the coast line with one or more islands), the previous setback code only computed the setbacks for a single geometry shape (which was essentially chosen at random). New version computes and combines setbacks for all geometries within a single county.

Updated baseline tests for new functionality, including some updates to the baseline test data.


grantbuster (approved):
Were we just not getting setbacks for the rest of the geometries then??

ppinchuk:
> Were we just not getting setbacks for the rest of the geometries then??

Yea... hard to say how much of an impact this had but once you knew what too look for, it was very noticeable for some counties, even in the RI test case.

grantbuster:
Ooof thats a big oops

PR #172, opened 2023-01-27
Reinforced Transmission

ppinchuk:
Reinforced Transmission
================
Add routines to calculate the least-cost-path transmission including reinforcement costs. More details about the methodology in the updated README.

Extra Features
=========
 - Fixed bug in `Geotiff` class regarding data shape
 - Added `min_line_length` as a user input, and set new default value to 0.
 - Added `poi_lat` and `poi_lon` to table outputs so they can be passed-through in reV (see [this reV issue](https://github.com/NREL/reV/issues/374)).


reVX/least_cost_xmission/least_cost_paths.py
```diff
@@ -380,3 +384,248 @@ def run(cls, cost_fpath, features_fpath, capacity_class,
...
+    balancing area network node.
+    """
+    def __init__(self, cost_fpath, features, transmission_lines,
+                 xmission_config=None):
+        """
+        Parameters
+        ----------
+        cost_fpath : str
+            Path to h5 file with cost rasters and other required layers.
+        features_fpath : str
```
grantbuster, line 401: Can you clarify this docstring? Is this supposed to be a filepath or a geodataframe? I think it get init from run() with a gdf?

reVX/least_cost_xmission/trans_cap_costs.py
```diff
@@ -1087,13 +1111,158 @@ def run(cls, cost_fpath, sc_point, features, capacity_class, radius=None,
...
+            corresponding existing transmission lines rastered into
+            them (i.e. array value is 1 at a pixel if there is a
+            transmission line, otherwise 0). These arrays will be used
+            to compute the reinforcement costs along existing
+            transmission lines of differing voltages.
+        cost_fpath : str
+            Full path of .h5 file with cost arrays.
+        start_idx : tuple
+            Tuple of (row_idx, col_idx) to compute reinforcement line
+            path to.
```
grantbuster, line 1159: This is just a general point of confusion on `start_idx` and `end_indices`. So these are the endpoint indexes in the `transmission_lines` rasters to compute cost paths between? Can we say "`(row_idx, col_idx) in the transmission_lines arrays`" (e.g. explicitly point back to the transmission lines input)? Both docstrings say "path to"... could we say "from start to end" or something similar so their relationship is explicit? How do we feel about naming them consistently, e.g. `start_indices` and `end_indices`?

reVX/least_cost_xmission/least_cost_paths.py
```diff
@@ -380,3 +384,248 @@ def run(cls, cost_fpath, features_fpath, capacity_class,
...
+        self._features, self._row_slice, self._col_slice, self._shape = \
+            self._map_to_costs(cost_fpath, features)
+        self._features = self._features.drop(columns='geometry')
+        self._cost_fpath = cost_fpath
+
+        network_node = (self._features.iloc[0:1]
+                        .dropna(axis="columns", how="all"))
+        err_msg = "Network node was dropped during clipping!"
+        assert "ba_str" in network_node, err_msg
+        self._start_idx = network_node[['row', 'col']].values[0]
```
grantbuster, line 431: Nice so this is the central BA network node that the substations will be reinforced to, right? How do you feel about adding a property that references this for documentation purposes since this is so different than the _start_idx of the main LCP code? Actually it could be useful to have properties for both start and end indices for LCP and RP:

```
class LCP:
@property 
def start_indices():
"""List of starting index values that index the starting points of transmission features in the raster array"""
indices = self.features.index if indices is None else indices
indices = [self._features.loc[start, ['row', 'col']].values for start in indices]
return indices

def end_indices():
"""List of ending index values that index the termination points of transmission features in the raster array, this list corresponds to self.start_indices and is the same length"""
# do something similar but with that self._clip_to_feature() call


class RP:
@property 
def start_indices():
"""Starting index of the transmission feature, for RP this is the single central BA network node"""
return self._start_idx

@property 
def end_indices():
"""List of ending index values of the transmission features to connect the starting index to. For RP this is a list of substation indexes that need to be reinforced to the start index (central BA network node)"""
return self._features[['row', 'col']].values

grantbuster (changes requested):
Looks pretty good. Great that you were able to leverage all of the previous LCP and TieLineCost work. Also thank you so much for updating the README!!!

reVX/least_cost_xmission/least_cost_paths.py
ppinchuk: Great catch! Updated below

reVX/least_cost_xmission/trans_cap_costs.py
ppinchuk: I think renaming to `start_indices` and `end_indices` is a good idea. Implemented that below. Also added quite a bit of elaboration in the docstrings as well. Let me know if there is still any confusion!

reVX/least_cost_xmission/least_cost_paths.py
ppinchuk: Ok, added those in! Let me know if you want to see any changes!

grantbuster (approved):
Nice i think this is a lot more clear, thanks for making those mods

Issue #170, opened 2023-01-25
Binary inclusions results


mjgleason:
**Bug Description**
`setbacks` module will sometimes produce binary results (i.e., exclusions) for counties with ordinances when inclusions should be returned (i.e., `weights_calculation_upscale_factor` is greater than 1).

Screen shot 1 shows an example of this occurring for roads setbacks in North Carolina. County boundaries are shown in black and the setbacks raster is displayed with a red/yellow/blue color ramp where `red=0` (full exclusion) and `blue=1` (full inclusion). Several counties show up with colors only of red and blue, indicating binary results. This was confirmed by inspecting the actual pixel values.

This same error occurs when there are no default setbacks used for counties without ordinances (see Screen shot 2). Interestingly, in this latter case, the resulting raster is actually of dtype `Byte` (i.e., `UInt8`) when it should be `float32`.

**Full Traceback**
There is no traceback since this is a semantic error.

**Code Sample**
NA

**To Reproduce**
To reproduce this problem, you can run `setbacks from-config -c this_config.json` where `this_config.json` is 
```json
{
  "execution_control": {
    "allocation": "setosa",
    "option": "eagle",
    "walltime": 1,
    "max_workers": 1,
    "memory": 180.0
  },
  "log_directory": "./logs",
  "log_level": "DEBUG",
  "excl_fpath": "/projects/rev/data/exclusions/ATB_Exclusions.h5",
  "regs_fpath": "/shared-projects/rev/projects/seto/data/solar_regulations_fix_2022-12-21.csv",
  "weights_calculation_upscale_factor": 5,
  "feature_type": "road",
  "features_path": "/projects/rev/data/conus/here_streets/Streets_USA_WY.gdb",
  "base_setback_dist": 0
}
```

**Expected behavior**
The expected result should be a raster is returned with inclusion values ranging from 0 to 1 instead of exclusion values of only 0 and 1 for these counties.

**Screenshots**
Screen shot 1:
![Screen Shot 2023-01-25 at 1 58 19 PM](https://user-images.githubusercontent.com/24549594/214690048-690fc8e0-1eb3-4364-98af-9e62943744b0.png)

Screen shot 2:
![Screen Shot 2023-01-25 at 2 03 56 PM](https://user-images.githubusercontent.com/24549594/214690276-e6129425-952a-4362-91eb-8cf8d251d860.png)

Screen shot 3:
[...]

PR #173, opened 2023-01-30
Fix Issue #170

mjgleason:
This PR implements the simpler fix for issue #170 proposed by @ppinchuk. I have tested this manually on the config provided in the #170 and it correctly produces an output inclusions raster. 

This PR supersedes PR #171 which I have closed without merging.


ppinchuk (approved):
Thanks @mjgleason For catching this and providing a fix, and especially for testing it on real setback data.

PR #180, opened 2023-04-27
Setbacks Refactor

ppinchuk:
This one is scary looking, but I think at least half of the modified files (and majority of modified lines) are test data files.

One thing we should definitely discuss before merging is how this will affect the conda release of `reVX`. At the moment I can't upload GAPs to the nrel conda channel because I don't have the correct token/permissions. Are you part of that org on anaconda @grantbuster? If so, maybe you can give me access so I can upload GAPs there. Otherwise it's possible conda won't be able to find the GAPs dependency (unless we don't list it I guess, but then users will have an extra step during their conda installation).

This all might not be that big of a deal since we encourage all uses to install with pip (like we do in the tests). If we stick to that, the above isn't really an issue.

Should go without saying but no rush on this one :)


reVX/setbacks/setbacks.py
```diff
@@ -0,0 +1,148 @@
...
+    if feature_filter_type not in FEATURE_FILTERS:
+        msg = ("Unknown feature filter type specified: {!r}. Must be one of {}"
+               .format(feature_filter_type, list(FEATURE_FILTERS)))
+        logger.error(msg)
+        raise ValueError(msg)
+
+    feature_subtypes_to_exclude = set(feature_subtypes_to_exclude or set())
+    num_features_per_worker = int(max(0, num_features_per_worker) or 10_000)
+
+    class Setbacks(AbstractBaseSetbacks):
```
grantbuster, line 84: Interesting syntax putting the class inside the function. I feel like there has to be a cleaner way to do this... Maybe consider having AbstractBaseSetbacks be non-abstract and decorated as a [dataclass](https://docs.python.org/3/library/dataclasses.html)?

Also add a return docstring, especially with such a funky output haha

reVX/setbacks/setbacks_cli.py
```diff
@@ -2,392 +2,552 @@
...
-    # direct command line instead of a config file for some bizarre reason
-    validate_setback_regulations_input(base_setback_dist, hub_height,
-                                       rotor_diameter)
+    for feature_name, feature_spec in (feature_specs or {}).items():
+        spec = SETBACK_SPECS.get(feature_name, {})
+        spec.update(feature_spec)
+        SETBACKS[feature_name] = setbacks_calculator(**spec)
+
+
+def compute_setbacks(excl_fpath, _ft, _fp, _mult, out_dir, tag,
```
grantbuster, line 220: I typically use a leading underscore to "hide" attributes of a class, what's the reason for underscore inputs like _ft and _fp? If this is a GAPS thing, any way we can actually hide these inputs as kwargs or something? Also, docstring does not match the actual variable names.

grantbuster (changes requested):
A few questions on structure, didn't dig too deep on the base setbacks class. 

Can you run an experiment where you simplify our [install code](https://github.com/NREL/reVX/blob/main/.github/workflows/pull_request_tests.yml#L31-L34)? Maybe try just this:

```
conda install pip pytest
pip install pytest-cov
pip install -e .
pip install HOPP

File: reVX/setbacks/setbacks.py (hunk 1 above)
Comment (ppinchuk) [reply reVX/setbacks/setbacks.py:84]: Yea interesting thought. I'm not sure how simple it would be to decorate with a dataclass since `AbstractBaseSetbacks` inherits from another abstract class. However, making it non-abstract is a really interesting idea. This function call is essentially a class factory, but I could move the factory logic into a `Setbacks` class method. That would make the syntax something like 
```
return Setbacks.new(feature_type, feature_subtypes_to_exclude, buffer_type, feature_filter_type, num_features_per_worker)
```

I quite like that actually. Would that address the issue in your opinion?

reVX/setbacks/setbacks_cli.py
ppinchuk: Yea essentially these are arguments that need to be explicit (they need to be detectable with `inspect.signature` - `**kwargs` would unfortunately obscure them) but shouldn't be exposed to the end-user documentation since they will always be provided by the code based on other user inputs.

I'm on the fence about this syntax. I think it's explicit and indicates the intent at a glance (as long as you are aware this is how GAPs works), but also it's kind of ugly and TBH probably doesn't belong in a public function. I think based on this feedback, I will explore ways to "hide" parameter names in the command setup class rather than relying on syntax. I think this idea should work - if not I will ping you to brainstorm other solutions

reVX/setbacks/setbacks.py
grantbuster: Well if the concrete setback class inherits the abstract class and just adds some class attributes, i would think we should make AbstractBaseSetbacks -> Setbacks where Setbacks is a data class (decorated with @dataclass or not). If all of the class attributes are required, just make them positional attrs. 

Class method that returns an instantiated Setbacks class would be a nice way to build and return the class. 

HMMM i guess is the main thing that I'm not understanding that you need the class returned and not an instance of the class?

ppinchuk: Right, this function is a class factory, so the return value is a class which can later be instantiated into an object. The reason this code is written this way is historical: In the past, each type of setback feature had it's own class implementation derived from a BaseSetback class. I wanted to put the power to generate new type of setback calculations into the users hands, so I consolidated the boilerplate of six child classes derived from BaseSetbacks into a single function that can generate the derived classes on demand.

I think it would technically be possible to move these attributes to the initializer. Again, the reason it wasn't done this way before is purely historical. Unfortunately this also means that making this change could be non trivial - the AbstractBaseSetbacks class would have to be renamed and refactored to allow extra initializer parameters, the parameter parsing logic would have to be moved somewhere within the class, and all code that instantiated/interacted with the derived setbacks classes (i.e. the majority of the setbacks tests) would have to be refactored to match the new init signature. This is all doable in theory, but I wonder if the effort is worthwhile in practice. If you think it is, please let me know and I can look into making this change.

In the meantime, I updated the syntax to be more functional in nature. Perhaps this alleviates some of the pain of the ugly old syntax?

reVX/setbacks/setbacks_cli.py
ppinchuk: We should definitely touch base as to why these "hidden" parameters are required. In the meantime, do you think he updated syntax suffices?

reVX/setbacks/setbacks.py
grantbuster: Sounds good, two things: camel case class name with strip, and ~reVX.setbacks.base.AbstractBaseSetbacks

ppinchuk:
I had to leave the `rtree` install in `conda install rtree pip pytest`, but other than that, the simplified install seems to work. Any final last minute comments/requests @grantbuster?

grantbuster:
Sweet that's great! glad geopandas fixed the shapely issue. 

Nope, merge away!

ppinchuk:
Wooo! Here goes nothing... :D

PR #182, opened 2023-05-23
ReEDS Supply Curve Column Additions CLI

ppinchuk:
Recently the ReEDS team has been asking for an extra set of fairly standardized supply-curve column outputs across many technologies (wind, solar, geothermal). Typically, the reV analysists pass around a script that adds these columns to the supply curve. However, changes to this script are hard to track and having a standardized utility for this (fairly simple) process could go a long way in terms of consistency in the outputs we pass to the ReEDS team. With that in mind, I added some utility functions to facilitate this process. The main entry point an analyst can use now is a new CLI command (powered by GAPs) that adds at least the following columns: "cnty_fips", "state", "county", "nrel_region", "eos_mult", and "reg_mult". More columns can be requested to be pulled from H5 files or JSON configs by the user.


reVX/utilities/reeds_cols.py
```diff
@@ -0,0 +1,274 @@
...
+        ``supply_curve_fpath``). By default, ``None``.
+    lat_col, lon_col : str, optional
+        Column names for latitude and longitude. These columns are
+        required to add county info. By default, ``"latitude"`` and
+        ``"longitude"``.
+    capacity_col : str, optional
+        Name of capacity column. This is used to filter out sites with
+        zero capacity, if that option is selected.
+        By default, ``"capacity"``.
+    extra_data : list of dicts
```
grantbuster, line 221: Are people manually preparing this input? might be better as a dataframe / csv input with two columns.

reVX/utilities/reeds_cols.py
```diff
@@ -0,0 +1,274 @@
...
+import pandas as pd
+from warnings import warn
+
+from rex import Resource
+from reVX.utilities.utilities import to_geo, load_fips_to_state_map
+from reVX.version import __version__
+
+COUNTY_GDF_FP = ("https://www2.census.gov/geo/tiger/TIGER2021/COUNTY/"
+                 "tl_2021_us_county.zip")
+NREL_REGIONS = {
```
grantbuster, line 17: you might end up getting a lot of these for various electrical system topologies (nemms regions, reeds ba's, etc...) You might consider putting this in a json in a folder of mappings. Assuming that somewhere we can just source a shape file too? There's a revx utility for doing the shape intersect but i'm not sure if you've included that here.

grantbuster (dismissed):
looks good but minor optional comments

reVX/utilities/reeds_cols.py
ppinchuk: It would be a manual preparation, yes. The format would look something like this:

```
extra_data = [
    {
        "data_fp": "./bespoke_out.h5",
        "dsets": ["ws_mean", "capital_cost", "cf_mean-2013"]
    },
    {
        "data_fp": "./sam.json",
        "dsets": ["total_losses"]
    }
]
```
Do you have any ideas about how to format this into a CSV? I can't think of anything clean since the `dsets` input is a variable-length list.

reVX/utilities/reeds_cols.py
ppinchuk: Yea good point about the JSON file for the mapping. I will add that.

What is the reVX utility you are referring to? The GDF file is used to do a spatial overlay with the input supply curve to extract county FIPS values.

reVX/utilities/reeds_cols.py
grantbuster: Gotcha, maybe not then, nevermind!

reVX/utilities/reeds_cols.py
ppinchuk: Interesting. I did play with it a bit when I started out but for some reason moved away from it. I just tried it again and it does seem to do something similar to what I am doing for the county lookups so I adjusted the logic to lean on that class. Thanks for the tip!

PR #183, opened 2023-06-05
Minor reeds col update

ppinchuk:
Allow rename mapping to happen before addition of `eos_mult`, `reg_mult`. These columns are expected by ReEDS and users shouldn't be renaming them anyways. On the flip side, it's entirely possible to rename a column into one of these, so this order makes a lot more logical sense.


ppinchuk:
@grantbuster This change is so minor that I might just merge it in once all tests are done

PR #184, opened 2023-06-07
Geotiff faster `lat_lon`

ppinchuk:
I got tired of waiting 93 minutes for a `lat_lon` calculation that takes < 45 seconds with vectorized code so I added that here. 
Still takes a while for large Geotiffs but it's much more manageable.

Also updated `rasterio` req, since newer version was causing some tests to fail. Haven't run full suite, so I will fix any remaining tests that fail due to the new dependecy


grantbuster (approved):
This is some magic gis stuff but looks like a good test so  👍

PR #187, opened 2023-07-21
Tw/reeds cols

WilliamsTravis:
There were potential problems in the remote county file read and the reprojection step (both potential network-related problem since pyproj uses a network connection too), so this tweak allows the user to pass their own local region file through add_reeds_cols. This also has the potential to significantly speed up the routine, depending on how the user treats this file.

I also turned off a CRS warning when finding the centroid for the outliers later. This might be a bit contentious, but since these are used to handle what is likely just handful of outliers (sc points that fall outside the boundaries of regions) it seems like a trivial warning. The best way would be to reproject the files into a local area projection, but that could take significantly more runtime and it would be difficult to find the most appropriate system for each region.

I also tweaked the function to add a dictionary option for the 'data_fp' argument in 'extra_data' so the user can do everything in one python script without having to write jsons.


reVX/utilities/reeds_cols.py
```diff
@@ -179,16 +188,17 @@ def add_reeds_columns(supply_curve_fpath, out_fp=None, capacity_col="capacity",
...
-        "dsets", and it points to a list of dataset names to extract
-        from the file. For JSON data extraction, the values of the
-        datasets must either be scalars or must match the length of the
-        input supply curve. For HDF5 data, the datasets must be 1D
-        datasets, and they will be merged with the input supply curve
-        on ``merge_col`` (column must be in the HDF5 file meta).
-        By default, ``None``.
+        The first key is "data_fp", and it points either a dictionary of
+        new field/new value pairs or to the path where the extra data is being
+        extracted from. This must be a dictionary, an HDF5, or JSON
```
ppinchuk, line 193: Can we update this text to something like

"The first key is "data_fp", and it points either a dictionary of new field/new value pairs or to the path where the extra data is being extracted from. The latter must be an HDF5 or JSON file (i.e., if not a dictionary, it must end in ".h5" or ".json")."

Basically I'm just trying to avoid saying "dictionary" so many times

reVX/utilities/reeds_cols.py
```diff
@@ -179,16 +188,17 @@ def add_reeds_columns(supply_curve_fpath, out_fp=None, capacity_col="capacity",
...
+        file (i.e., if not a dictionary, it must end in ".h5" or ".json"). The
+        second key is "dsets", and it points to a list of dataset names to
+        extract from data_fp. For JSON and dictionary data extraction, the
+        values of the datasets must either be scalars or must match the length
+        of the input ``data_frame``. For HDF5 data, the datasets must be 1D
+        datasets, and they will be merged with the input ``data_frame``
+        on ``merge_col`` (column must be in the HDF5 file meta). By default,
+        ``None``.
```
ppinchuk, line 201: Can you add documentation for the `regions` input parameter at the end of this docstring? It doesn't have to be anything long or fancy, but this is the text that gets put into the documentation for this function, so it's important that we have all of our parameters documents.

reVX/utilities/reeds_cols.py
```diff
@@ -127,10 +133,12 @@ def add_extra_data(data_frame, extra_data, merge_col="sc_point_gid"):
         dsets = data_info["dsets"]
         data_fp = data_info["data_fp"]
 
-        if data_fp.endswith(".json"):
+        if isinstance(data_fp, dict):
+            extra_data = data_fp
+        elif str(data_fp).endswith(".json"):
```
ppinchuk, line 138: Is this cast to `str` (and the one below) really necessary? This value should either be a string or a dictionary, and we already check for a dictionary above. Any other kind of input *should* error out here.

ppinchuk (changes requested):
Great little change! Always love additional flexibility for end-users.
Just a few comments to help up keep our documentation clean and up-to-date.
Thanks for this, Travis!

reVX/utilities/reeds_cols.py
WilliamsTravis: It could be a Posix Path, can add to the docs.

ppinchuk: Ah gotcha. I don't think you need to update the docs then, unless you really want to

ppinchuk:
@WilliamsTravis I went ahead and made some of the requested updates and then merged code to fix the tests. Once they all pass, this branch should be good to merge.

One thing I ended up doing is changing the dictionary key name from `data_fp` to `source`. This will show up in the updated documentation, but would be good to be aware of it for your config files

ppinchuk (approved):

PR #191, opened 2023-08-09
Make slope friction parameters user changable

mikebannis:


ppinchuk (approved):
I went ahead and converted the constants to uppercase to stick with convention (and make it immediately obvious in the code that these are constant values).

I also converted some rogue print statements to logger calls.

reVX/least_cost_xmission/aoswt_utilities.py
```diff
@@ -100,15 +100,15 @@ class CombineRasters:
     LAND_MASK_FNAME = 'land_mask.tif'
 
     # Slope cutoffs (degrees)
-    slope_barrier_cutoff = 15  # slopes >= this value are for barriers & high
+    SLOPE_BARRIER_CUTOFF = 15  # slopes >= this value are for barriers & high
```
mikebannis, line 103: These are not supposed to be constant, but attributes that can be changed by the consumer code.

ppinchuk: I guess not constant in the literal sense but more like representing a number. If these are going to be changed frequently, they should probably be moved to the `__init__`. Otherwise they are "constant enough" to warrant uppercase :p

reVX/least_cost_xmission/least_cost_xmission_cli.py
```diff
@@ -148,7 +147,7 @@ def from_config(ctx, config, verbose):
               show_default=True, default=None,
               help=("Number of workers to use for processing, if 1 run in "
                     "serial, if None use all available cores"))
-@click.option('--out_dir', '-o', type=STR, default='./',
+@click.option('--out_dir', '-o', type=STR, default='./out',
```
ppinchuk, line 150: We should probably add a call to `create_dirs` in the function body below to make sure `./out` exists

reVX/least_cost_xmission/least_cost_xmission_cli.py
```diff
@@ -229,47 +228,47 @@ def local(ctx, cost_fpath, features_fpath, balancing_areas_fpath,
...
                 logger.info('--drop options must on or more of {}, received {}'
                             .format(TRANS_CAT_TYPES, drop))
                 return
 
-    logger.info('Loading: {}'.format(", ".join(files)))
     warnings.filterwarnings('ignore', category=RuntimeWarning)
-    df = pd.concat([gpd.read_file(f) if "gpkg" in f else pd.read_csv(f)
-                    for f in files])
+    dfs = []
+    for i, file in enumerate(files):
```
ppinchuk, line 267: can you change this to  `... enumerate(files, start=1):`? That way we get (1/10) .. (10/10) instead of (0/10) .. (9/10) in the logger string

ppinchuk (changes requested):
Just a couple minor requests, since you'll have to battle the linter anyways :P

reVX/least_cost_xmission/least_cost_xmission_cli.py
mikebannis: Love it, will do

reVX/least_cost_xmission/least_cost_xmission_cli.py
mikebannis: 👍

ppinchuk (approved):

PR #193, opened 2023-08-17
Doc/document new aoswt options

mikebannis:


ppinchuk (approved):
Thanks for doing this, the notebook looks great!

PR #196, opened 2023-09-20
Changing sc_gid to sc_point_gid to match supply curve data across different datasets

pesap:
## Purpose of this PR

- Changing `sc_gid` to `sc_pointg_gid` for the nodal Plexos aggregation,
- Adding updated column names from the supply curve files and profiles datasets.

##  Background

We experienced some difficulties when matching `sc_gid` for the latest NTP runs that use the latest reV runs. After a discussion with @mmowers and @WilliamsTravis, we found out that we should be using `sc_point_gid` instead of `sc_gid` to match across supply curves and profiles datasets that are used for the nodal translation.

I added this PR on this branch since we end-up using the updated bespoke wind profiles as well. Is there a timeline for your branch to be merged into main, @grantbuster ?


grantbuster:
@pesap, your branch is broken, please rebase before we talk about this. See this for details: https://github.com/NREL/reVX/network

pesap:
> @pesap, your branch is broken, please rebase before we talk about this. See this for details: https://github.com/NREL/reVX/network

@grantbuster, It looks like it is rebased now. Although, I am not sure why there are so many commits on it. I do see the my changes only in the changed files. How does it looks on your end?

grantbuster:
> > @pesap, your branch is broken, please rebase before we talk about this. See this for details: https://github.com/NREL/reVX/network
> 
> @grantbuster, It looks like it is rebased now. Although, I am not sure why there are so many commits on it. I do see the my changes only in the changed files. How does it looks on your end?

It looks horribly messed up on my end. You're duplicating commits and this will result in a really crazy git history.

pesap:
I am going to close this PR and will do a separate PR with my own fork.

PR #197, opened 2023-09-20
Changing sc_gid to sc_point_gid for matching gid points

pesap:
## Purpose of this PR

- Changing `sc_gid` to `sc_pointg_gid` for the nodal Plexos aggregation,
- Adding updated column names from the supply curve files and profiles datasets.

##  Background

We experienced some difficulties when matching `sc_gid` for the latest NTP runs that use the latest reV runs. After a discussion with @mmowers and @WilliamsTravis, we found out that we should be using `sc_point_gid` instead of `sc_gid` to match across supply curves and profiles datasets that are used for the nodal translation.

I added this PR on this branch since we end-up using the updated bespoke wind profiles as well. Is there a timeline for your branch to be merged into main, @grantbuster ?

### DIFF
--- a/reVX/plexos/rev_reeds_plexos.py
+++ b/reVX/plexos/rev_reeds_plexos.py
@@ -311,7 +311,7 @@ class PlexosAggregation(BaseProfileAggregation):
         reeds_build : pd.DataFrame
             Same as input but without lat/lon columns if matched.
         """
-        join_on = 'sc_gid'
+        join_on = 'sc_point_gid'
         reeds_build = reeds_build.sort_values(join_on)
         reeds_sc_gids = reeds_build[join_on].values
         rev_mask = rev_sc[join_on].isin(reeds_sc_gids)
@@ -391,8 +391,8 @@ class PlexosAggregation(BaseProfileAggregation):
 
         reeds_build = reeds_build[year_mask]
 
-        join_on = 'sc_gid'
-        if 'sc_gid' not in rev_sc or 'sc_gid' not in reeds_build:
+        join_on = 'sc_point_gid'
+        if 'sc_point_gid' not in rev_sc or 'sc_point_gid' not in reeds_build:
             raise KeyError('GID must be in reV SC and REEDS Buildout tables!')
 
         rev_sc, reeds_build = cls._check_rev_reeds_coordinates(rev_sc,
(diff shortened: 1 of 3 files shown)


grantbuster:
@pesap thanks for opening a clean PR. 

Can ReEDS still build infrastructure based on multiple resource classes? That is the main reason we have both sc_gid and sc_point_gid. In order to maintain this functionality I recommend adding a kwarg to set the gid column instead of just hard-coding.

pesap:
> @pesap thanks for opening a clean PR.
> 
> Can ReEDS still build infrastructure based on multiple resource classes? That is the main reason we have both sc_gid and sc_point_gid. In order to maintain this functionality I recommend adding a kwarg to set the gid column instead of just hard-coding.

I think so ( @mmowers  can correct me if I am wrong). Sure, we can just pass the gid_column as kwarg.

pesap:
@grantbuster , is this how you were envisioning this? Happy to modify or adapt it to the reVX flavor.

reVX/plexos/rev_reeds_plexos.py
```diff
@@ -311,7 +318,7 @@ def _check_rev_reeds_coordinates(rev_sc, reeds_build, atol=0.5):
         reeds_build : pd.DataFrame
             Same as input but without lat/lon columns if matched.
         """
-        join_on = 'sc_gid'
+        join_on = gid_column
```
grantbuster, line 321: probably an unnecessary variable assignment?

reVX/plexos/rev_reeds_plexos.py
```diff
@@ -391,8 +398,8 @@ def _parse_rev_reeds(cls, rev_sc, reeds_build, build_year=2050,
 
         reeds_build = reeds_build[year_mask]
 
-        join_on = 'sc_gid'
-        if 'sc_gid' not in rev_sc or 'sc_gid' not in reeds_build:
+        join_on = gid_column
```
grantbuster, line 401: probably an unnecessary variable assignment?

reVX/plexos/rev_reeds_plexos.py
```diff
@@ -444,9 +451,9 @@ def _check_gids(self):
             else:
                 gid_col = list(gid_col)
 
-            for i, sc_gids in enumerate(gid_col):
-                if any(m in sc_gids for m in missing):
-                    bad_sc_points.append(self.sc_build.iloc[i]['sc_gid'])
+            for i, sc_point_gid in enumerate(gid_col):
+                if any(m in sc_point_gid for m in missing):
```
grantbuster, line 455: I think this variable should still be called "sc_gids"

grantbuster (changes requested):
Looks good, few minor edits and then if the tests pass we can merge

grantbuster (approved):

grantbuster:
@pesap looks good, thanks for making those changes! let tests run and then merge please.

PR #198, opened 2023-09-27
LCP QOL updates

ppinchuk:
Add logic to submit futures in chunks of `max_workers` for LCP calculations. This ensures that user can control the memory consumption of the LCP computation by setting `max_workers` (i.e. futures that don't start processing right away won't eat up a whole node's worth of memory during submission). No new tests required for this change, since we had tests for parallel LCP computation already.

Also a added a `clip_buffer` option to address a very minor bug where the LCP sometimes cannot be found if the clipping process creates islands. The new tests demonstrates this instance. 

Also fixes HOPP dependency that breaks our tests by pinning HOPP version


mikebannis (approved):
LGTM

PR #199, opened 2023-10-05
Make safe filenames from POIs

mikebannis:


ppinchuk (approved):

ppinchuk:
Actually, this is so minor that we can merge without waiting

PR #210, opened 2024-03-26
Feat/track dry vs wet costs

mikebannis:
Allow costs and lengths for sub layers (cost layers) to be determined individually. Additionally:

- Landfall costs are unaffected by path direction. Diagonal paths typically have an increased cost due to the longer route.
- When cost layers are specified, the total cost for the path is calculated using the cost layers, not the combined cost raster. This ensure that landfall costs are calculated properly.


reVX/utilities/exceptions.py
```diff
@@ -16,6 +16,10 @@ class ExclusionsCheckError(Exception):
     """
 
 
+class ProfileCheckError(ValueError):
+    """Profile mismtahc between two layers. """
```
mikebannis, line 20: spelling

reVX/least_cost_xmission/layers/transmission_layer_io_handler.py
```diff
@@ -240,18 +190,21 @@ def load_tiff(self, fname: str, band: int = 1,
...
-                                 f'({self.shape}).')
-            if transform != self.profile['transform']:
-                raise ValueError(f'Transform of {full_fname}:\n{transform}\n'
-                                 'does not match template raster shape:\n'
-                                 f'{self.profile["transform"]}')
-            if crs != self.profile['crs']:
-                raise ValueError(f'CRS of {full_fname}:\n{crs}\ndoes not '
-                                 'match template raster shape:\n'
-                                 f'{self.profile["crs"]}')
+            pass
```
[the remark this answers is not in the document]
ppinchuk: Yea sorry, still actively working on this PR. I don't think I can convert it to a draft since it's your PR but I can request a review from you once it's ready.

It's going to be a big boy PR again lol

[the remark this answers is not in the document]
ppinchuk: P.S. TO actually answer your question - this part of the code was erroring out when it shouldn't be, so I am going to replace this with the more nuanced checks we had in `ExclusionsConverter` (which is now called `LayeredH5`)

ppinchuk:
@mikebannis I think this is ready for review now

reVX/config/least_cost_xmission.py
```diff
@@ -5,6 +5,7 @@
 import os
 import logging
 import pandas as pd
+from typing import List
```
mikebannis, line 8: I believe `DryCostCreatorConfig` below is no longer used

reVX/config/transmission_layer_creation.py
```diff
@@ -80,13 +82,56 @@ class WetCosts(BaseModel, extra='forbid'):
...
+        - "land_use_classes"
+        - "new_substation_costs"
+        - "power_classes"
+        - "power_to_voltage"
+        - "transformer_costs"
+        - "upgrade_substation_costs"
+
+    Each of these keys should point to a dictionary or a path to
+    a separate json file contianing a dictionary of
+    configurations for each section."""
```
mikebannis, line 115: This doc string is in the repo three times. Converting the `XmissionConfig` class into Pydantic would reduce the need for the doc strings and define the allowed keys more legibly in code, vs in doc. But, it's also more time and effort.

reVX/config/transmission_layer_creation.py
```diff
@@ -80,13 +82,56 @@ class WetCosts(BaseModel, extra='forbid'):
...
+
+    default_mults: Optional[IsoMultipliers] = None
+    """Multipliers to be used for default region.
+
+    This input should be a dictionary with three keys:
+
+        - "iso": Thie key is ignored, but is required. Can set to
```
mikebannis, line 122: A required but ignored key, that sounds like I screwed up. Can we pull it entirely?

reVX/least_cost_xmission/costs/landfall_cost_creator.py
```diff
@@ -0,0 +1,46 @@
...
+
+class LandfallCostCreator(LayerCreator):
+    """
+    Create landfall costs and save to GeoTIFF.
+    """
+
+    def build(self, landfall_cost: float,
+              landfall_layer_name: str = LANDFALL_COSTS_H5_LAYER_NAME):
+        """
+        Build complete offshore costs. This is currently very simple. In the
```
mikebannis, line 24: Docstring appears copy-pasted from offshore costs and is somewhat erroneous

reVX/least_cost_xmission/layers/friction_barrier_builder.py
```diff
@@ -7,58 +7,61 @@
...
+            Directory where cost layers should be saved as GeoTIFF.
+            By default, ``"."``.
         dtype : np.dtype, optional
             Data type for final dataset. By default, ``float32``.
         """
-        self._type = type_
-        self._io_handler = io_handler
-        self._masks = masks
-        self._dtype = dtype
+        super().__init__(io_handler=io_handler, mask=mask,
```
mikebannis, line 46: docstring for `mask` states `ndarray` in `LayerCreator` may be out of data. Passing a `Masks` instance here

mikebannis: Thinking about this more, I'm tempted to say make `mask` optional in `LayerCreator`, and create a `self._masks: Masks` attribute just for `FrictionBarrierBuilder` so there is no confusion. Or get a little fancier and allow both on `LayerCreator` but FBB seems to be the only subclass that needs it.

reVX/least_cost_xmission/trans_cap_costs.py
```diff
@@ -260,42 +258,74 @@ def _parse_config(xmission_config=None):
...
+        logger.debug("Building cost layer with the folowing inputs:"
+                     f"\n\t- cost_layers: {cost_layers}"
+                     f"\n\t- length_invariant_cost_layers: {li_cost_layers}"
+                     f"\n\t- barrier_mult: {barrier_mult}")
+        self._cost = np.zeros(self.clip_shape, dtype=np.float32)
+        overlap = np.zeros(self.clip_shape, dtype=np.uint8)
+        with ExclusionLayers(self._cost_fpath) as f:
+            for cost_layer in cost_layers:
+                cost = f[cost_layer, self._row_slice, self._col_slice]
+                self._cost += cost
```
mikebannis, line 291: @ppinchuk , do you think there is a significant performance hit for doing this every time?

reVX/least_cost_xmission/trans_cap_costs.py
```diff
@@ -260,42 +258,74 @@ def _parse_config(xmission_config=None):
...
+
+            for li_cost_layer in li_cost_layers:
+                li_cost = f[li_cost_layer, self._row_slice, self._col_slice]
+                overlap += li_cost > 0
+                self._li_cost_layer_map[li_cost_layer] = li_cost
+
+            barrier = f[self.TB_LAYER_NAME, self._row_slice, self._col_slice]
+            barrier = barrier * barrier_mult
+
+        if (overlap > 1).any():
```
mikebannis, line 303: Love it

reVX/least_cost_xmission/trans_cap_costs.py
```diff
@@ -926,6 +1033,11 @@ def compute_tie_line_costs(self, min_line_length=5.7,  # noqa: C901
         if save_paths:
             paths = []
 
+        if cost_layers is not None:
```
mikebannis, line 1036: Is `cost_layers` still optional with the new methodology?

reVX/least_cost_xmission/trans_cap_costs.py
```diff
@@ -1330,3 +1488,21 @@ def run(cls, transmission_lines, cost_fpath, start_indices, end_indices,
                      .format((time.time() - ts) / 60))
 
         return tie_lines
+
+
+def _compute_individual_layers_costs_lens(layer_map, indices, lens, results,
+                                          scale_by_length=True):
+    """Compute costs and dists by layer. """
```
mikebannis, line 1495: Params/return in docstring and/or types please

mikebannis (commented):
Generally looks good. Left some comments, apologies for the nit-picks!

reVX/config/transmission_layer_creation.py
ppinchuk: Just looked and I think we can! I wasn't sure if you had some purpose for it or not so that's why i left it, but I will go ahead and remove it now. Good catch!

reVX/least_cost_xmission/costs/landfall_cost_creator.py
ppinchuk: Oops, good catch. Will update

reVX/least_cost_xmission/trans_cap_costs.py
ppinchuk: It should not be significant, since we are only reading in a small portion of the array.

But even if it was, what do you think would be a good alternative here? If we read in the whole array, we may end up dealing with even more memory issues than we are already. As such, I am happy to trade a little bit of CPU time to save memory. Thoughts on that?

reVX/least_cost_xmission/layers/friction_barrier_builder.py
ppinchuk: Yea, this was the trickiest part - I tried to get rid of the `Masks` dependency for FBB but couldn't do it. BUT, this is the only difference between FBB and the `LayerCreator` base, so I still went with it.

I do think your idea of making it optional may be the best bet, so I will do that!

reVX/least_cost_xmission/trans_cap_costs.py
mikebannis: Sounds good to me, I briefly forgot about the MCP memory bottleneck

reVX/least_cost_xmission/layers/friction_barrier_builder.py
mikebannis:
```
class FrictionBarrierBuilder(LayerCreator):
    def __init__(self,  masks: Masks, ...):
        super().__init__(io_handler=io_handler,
                         output_tiff_dir=output_tiff_dir, dtype=dtype)
        self._masks: Masks = masks
```

mikebannis: Something like that? possible confusion between `_mask` and `_masks`, but I'm ok with it.

reVX/least_cost_xmission/trans_cap_costs.py
ppinchuk: Nope, good catch. I will fix this

reVX/least_cost_xmission/layers/friction_barrier_builder.py
ppinchuk: yup, that's pretty much exactly what I threw in there!

ppinchuk (approved):
Excited to get this functionality merged!

PR #213, opened 2024-05-10
Generalize Layer Creator Functionality

ppinchuk:
@mikebannis feel free to take a peek and let me know if you see any red flags (but no pressure)

This PR gets about 90% of the way toward generalizing the Layer Creator class. What is left to do is move it to a more general location and build a small CLI for it, but that can happen once we get a project that needs this kind of functionality


reVX/config/transmission_layer_creation.py
```diff
@@ -44,20 +45,37 @@ class Rasterize(BaseModel, extra='forbid'):
...
+    Friction and barrier layers config model.
+    'global_value', 'map', 'bins', 'rasterize', and 'forced_inclusion'
+    are exclusive, but exactly one must be specified. Example configs
+    can be seen in `test_xmission_barrier_friction_builder.py` in the
+    tests directory.
     """
-    extent: Extents
-    """Extent to apply map or range to"""
+
+    extent: Optional[Extents] = ALL
```
mikebannis, line 57: I believe `Optional` implies `None` is an option. This looks like it's just `Extents` with a default

reVX/config/transmission_layer_creation.py
```diff
@@ -140,27 +149,71 @@ class MergeFrictionBarriers(BaseModel, extra='forbid'):
     by a factor. The multiplier should be large enough that all barriers have
     a higher value than any possible friction.
     """
+
+    friction_layer: str
```
mikebannis, line 153: Should these be `FilePath`? That may have issues if the file doesn't exist when the script is started

reVX/config/transmission_layer_creation.py
```diff
@@ -140,27 +149,71 @@ class MergeFrictionBarriers(BaseModel, extra='forbid'):
...
 
-BarrierLayers = Dict[str, FBLayerConfig]
-"""Mapping of barrier layers.
+    description: Optional[str] = None
+    """Optional description to store in attrs for layer. """
 
-Keys are GeoTIFF or vector filepaths. Values are the FBLayerConfig to use for
-that file."""
+    include_in_h5: Optional[bool] = True
+    """Flag to specify wether layer should be stored in H5 or not. """
```
mikebannis, line 196: whether

reVX/least_cost_xmission/layers/layer_creator.py
```diff
@@ -130,7 +168,7 @@ def _process_raster_layer(self, data: npt.NDArray,  # type: ignore[return]
 
             return processed
 
-        # No range, has to be map. Assign cells values based on map.
+        # No bins specified, has to be map. Assign cells values based on map.
```
mikebannis, line 171: Is it possible to end up here by user error when map is not intended?

mikebannis:
Generally looks good, left a couple minor comments.

mikebannis (approved):

reVX/config/transmission_layer_creation.py
ppinchuk: Oops, didn't know that! Will fix.

reVX/config/transmission_layer_creation.py
ppinchuk: These guys are just the layer names - the path is assumed to be `<friction_layer>.tif`  if it exists on disk. This is documented in the docstring, and the reason I leave it flexible is so that you can create the friction layer on the fly without having it on disk when you launch the script.

Let me know if you see any issues with that

reVX/least_cost_xmission/layers/layer_creator.py
ppinchuk: Yea good question. Let's see...

`forced_inclusion is not None` is handled outside of this function, as is `rasterize is not None`.

`global_value is not None` returns within this function above, and so does `bins is not None` does the same.

I think the only possible issue here may be if none of the required config keys are set, but the CLI checks for this. So as long as you are always using the CLI, this should be perfectly ok. 

Anything I miss?

reVX/config/transmission_layer_creation.py
mikebannis: 👍

ppinchuk:
Thanks for the review @mikebannis - these were exactly the kinds of things I wanted to catch :)

PR #224, opened 2024-07-02
Gb/plexos bespoke

grantbuster:


grantbuster:
@ppinchuk this PR hacks the bespoke supply curve to allow for reVX utilities to build "plant" generation profiles from bespoke generation files. Typically we go through a supply curve and build a plant profile based on included area per nsrdb/wtk resource gid. For bespoke, each supply curve point has its own profile and is not an aggregation of profiles at the resource level.

grantbuster:
Also 100% no rush on this... it's sat for a year it can sit for a little longer haha. At least we have an open PR now!

reVX/plexos/utilities.py
```diff
@@ -99,6 +99,9 @@ class DataCleaner:
 
     REV_NAME_MAP = {'gid': 'sc_gid',
```
ppinchuk, line 100: You went through a great deal of effort in the PR to add a `gid_column` input to many functions that defaults to `"sc_gid"`, but it look slike you never actually use anything other than the default value of `"sc_gid"`. Do you set it to something else in your personal scripts? Or does it just work because of this mapping? If its the latter, do we actually need the `gid_column` input in all of those functions?

ppinchuk (commented):
I definitely don't have a full grasp on what is going on here. It's a little wild to me to be setting `res_gids` to `sc_point_gid` (or equivalent), but I definitely won't object.

If you need to get this merged ASAP then go for it, my one recommendation would be to be extra sure its actually doing what you want to - mainly pulling in the profiles you want to be pulling

ppinchuk:
Oh and also I think this code may break with reV 0.9+ bespoke outputs. In particular `total_lcoe` has a way different name now, as do some other columns (not sure how extensive the rev rename map is)

reVX/handlers/sc_points.py
```diff
@@ -14,6 +14,7 @@
 from rex.resource import Resource
 from rex.utilities import parse_table, SpawnProcessPool
 
+from reVX.plexos.base import BaseProfileAggregation
```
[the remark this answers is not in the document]
grantbuster: unused... not sure why added...

reVX/plexos/base.py
```diff
@@ -705,6 +716,38 @@ def get_unique_plant_names(table, name_col, tech_tag=None):
...
+
+        Returns
+        -------
+        table : pd.DataFrame
+            Modified so that the "res_gids" points to the supply curve gids in
+            the bespoke file so that the resource generation profiles will be
+            taken based on the bespoke indexing.
+        """
+
+        df['res_gids'] = [[g] for g in df[gid_column]]
```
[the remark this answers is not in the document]
grantbuster: So without bespoke we lookup the sc point -> res_gids -> res_gid in generation file in order to build the profile. With bespoke what we're trying to do is map the SC point to the bespoke index. Everything in this case is sc_gid. 

I think i see your point that gid_column really NEEDS to be sc_gid if we're using bespoke. will add an exception here, but we'll still let gid_column be other stuff if not bespoke.

reVX/plexos/utilities.py
grantbuster: ugh i dont remember! i think i was just trying to hardcode less of this stuff but i agree this seems unnecessary. Maybe will remove and hardcode?

grantbuster: oh i see, note from pesap: "For the Plexos Nodal aggregation we are going to pass
the gid_column as "sc_point_gid""

https://github.com/NREL/reVX/pull/224/commits/d689c602e17fcd1b7eba53688530b14aac48f7da

grantbuster:
> Oh and also I think this code may break with reV 0.9+ bespoke outputs. In particular `total_lcoe` has a way different name now, as do some other columns (I think `REV_NAME_MAP` covers this issue to a degree but not sure how extensive that solution is currently)

If the tests pass let's merge and fix this ad-hoc.

